PR #17, opened 2018-09-27
fix: drop top-level from path

darscan:
- [x] Ready for review

#### What does this PR do?

- drops the top-level `from` attribute from the response (so that we know to annotate it later)


michael-go:
oh noes - sorry! missed this in https://github.com/snyk/snyk-gradle-plugin/pull/10

Issue #22, opened 2018-11-13
Gradle, if annotationProcessor is used then only those dependencies are being analysed


Vortim:
- `node -v`: 9.11.1
- `npm -v`: 5.10.0
- `snyk -v`: 1.108.2
- Command run: snyk test

### Expected behaviour
Expecting snyk to ignore annotationProcessor dependencies **or** add those dependencies to others found in project.

### Actual behaviour
Only annotationProcessor dependencies are being analysed (in my case it's only lombok)

### Steps to reproduce
Simply execute `snyk test` in project root directory.

### Additional info
In my humble opinion the root cause of issue is in filters (that only first 'configuration' is being used for analysis): https://github.com/snyk/snyk-gradle-plugin/blob/9d0692e5389afa31eaf42bd2d3385b2586bda380/lib/gradle-dep-parser.js#L24

Sample ./gradlew dependencies output:

```
> Task :dependencies

------------------------------------------------------------
Root project
------------------------------------------------------------

annotationProcessor - Annotation processors and their dependencies for source set 'main'.
\--- org.projectlombok:lombok -> 1.18.2

apiElements - API elements for main. (n)
No dependencies

archives - Configuration for archive artifacts.
No dependencies

bootArchives - Configuration for Spring Boot archive artifacts.
No dependencies

compile - Dependencies for source set 'main' (deprecated, use 'implementation' instead).
No dependencies

compileClasspath - Compile classpath for source set 'main'.
+--- com.github.ben-manes.caffeine:caffeine -> 2.6.2
+--- com.github.kstyrc:embedded-redis -> 0.6
|    +--- com.google.guava:guava:18.0 -> 25.0-jre
|    |    +--- com.google.code.findbugs:jsr305:1.3.9 -> 3.0.2
|    |    +--- org.checkerframework:checker-compat-qual:2.0.0
|    |    +--- com.google.errorprone:error_prone_annotations:2.1.3
|    |    +--- com.google.j2objc:j2objc-annotations:1.1
|    |    \--- org.codehaus.mojo:animal-sniffer-annotations:1.14
|    \--- commons-io:commons-io:2.4 -> 2.6
+--- com.google.code.findbugs:jsr305 -> 3.0.2
```

---

### --debug output
```
  snyk test { _: [ [Circular] ], debug: true } +0ms
  snyk no file specified. Trying to autodetect in base folder <path_to_base_project_dir> +0ms
[...]

Issue #19, opened 2018-10-18
wrong behavior when include checkstyle plugin


quynh-axiadids:
- `node -v`: v9.7.1
- `npm -v`: 5.6.0
- `snyk -v`: 1.104.1
- `gradle -v`: 4.8
- build.gradle
```
apply plugin: 'java'
apply plugin: 'checkstyle'

ext {
    checkstyleVersion = '8.10.1'
    jacksonVersion = '2.9.7'
    springsecuritysamlVersion = '1.0.3.RELEASE'
}

group 'com.test'
version '0.8.6'

sourceCompatibility = 1.8

repositories {
    mavenCentral()
}

dependencies {
    implementation group: 'com.fasterxml.jackson.core', name: 'jackson-annotations', version: jacksonVersion
    implementation group: 'org.springframework.security.extensions', name: 'spring-security-saml2-core', version: springsecuritysamlVersion
}
```
- Command run: snyk test

### Expected behaviour
```
Organisation:      <>
Package manager:   gradle
Target file:       build.gradle
Open source:       no
Project path:      <>

Tested 33 dependencies for known vulnerabilities, found 26 vulnerabilities, 96 vulnerable paths.
```

### Actual behaviour
```
✗ Medium severity vulnerability found in com.google.guava:guava
  Description: Deserialization of Untrusted Data
  Info: https://snyk.io/vuln/SNYK-JAVA-COMGOOGLEGUAVA-32236
  Introduced through: com.puppycrawl.tools:checkstyle@6.19
  From: com.puppycrawl.tools:checkstyle@6.19 > com.google.guava:guava@19.0

Organisation:      <>
Package manager:   gradle
Target file:       build.gradle
Open source:       no
Project path:      <>

Tested 7 dependencies for known vulnerabilities, found 1 vulnerability, 1 vulnerable path.
```

After remove `apply plugin: 'checkstyle'`, the snyk test gives expects results.
It seems that checkstyle plugin causes snyk to behave wrongly and run test on the checkstyle dependencies instead.

---

If applicable, please append the `--debug` flag on your command and include the output here **ensuring to remove any sensitive/personal details or tokens.

adrukh:
@ quynh-axiadids thanks for reporting this!

Your analysis is very much correct. We're thrown off by the `checkstyle` plugin as it introduces its own configuration to the gradle build, and pick it up instead of the rest of the app.

[...]

PR #23, opened 2018-11-23
Feat/default configuration

adrukh:
- [x] Ready for review

#### What does this PR do?

* Tidies up `semantic-release` flow, removes `semantic-release` as a devDep
* Runs an extra `gradle` command to list the plugins in the analyzed project
* If the `'java'` gradle plugin is included in the analyzed project, AND no explicit gradle configuration is set via `options.args`, then the `default` gradle configuration will be used for our analysis.

Why? Good question.

Our current logic parses the dependencies of the *first* configuration with any dependencies out of the list of all relevant configurations as they are printed by the `gradle dependencies` command. This logic is quite arbitrary and causes some silent mistakes for our users, see #19 and #22 for example.

A user pointed out that the `default` gradle configuration would be more sensible as a fallback. We believe that the `default` gradle configuration is only valid with the `'java'` plugin, hence this PR.

We're totally open to other approaches and would like to crowd-source them from our savvy gradle users. Calling on @quynh-axiadids, @Vortim as well as @pastjean, would be very grateful for your input here.

An alternative approach for implicit configuration selection was discussed internally @snyksec, and it looks like this:
1. Run `gradle dependencies` like we do today, collect output of all relevant configurations
2. Take the following configurations only, in the exact order as they appear here:
```
runtimeClasspath
runtime
compileClasspath
compile
default
```
3. Filter out those that have no dependencies
4. Analyze the *first* of the remaining configurations.

We'd appreciate any comments! 🙏

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,9 @@ script:
 jobs:
   include:
     - stage: npm release
+      if: branch = master AND type != pull_request
       node_js: "8" # This *has* to be the "build leader"
-      script: skip
-      after_success:
-        - npm run semantic-release
+      script: npx semantic-release
 branches:
   only:
     - master
--- a/lib/gradle-plugin-parser.js
+++ b/lib/gradle-plugin-parser.js
@@ -1,0 +1,21 @@
+module.exports = {
+  parse: parse,
+};
+
+function parse(text) {
+  var pluginRe = /^[a-zA-Z0-9_\-\.]+@[0-9a-f]+$/;
+  if (text && text.length) {
+    return text.split('\n')
+      .map(trim)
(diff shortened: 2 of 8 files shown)


pastjean:
I wouldnt say i'm gradle savy. But you need to make sur the test dependencies are also included with the "dev" flag

Also in gradle v5 dependency usage changed around a little with the "api" and "implementation" instead of "compile" target so I dont know how this affects resolution while building in classpaths

lili2311:
👋 Just an update that we are closing this PR in favour of https://github.com/snyk/snyk-gradle-plugin/pull/29, we spent some time planning this feature last week and are now ready to start rolling this out.

PR #24, opened 2018-12-08
Fix/path with space

adrukh:
- [x] Ready for review

#### What does this PR do?

The plugin failed to process a project whose target file path had spaces. That is `inspect('.', 'path with space/to/build.gradle, options)` would fail.

Fixing this by quoting the target file path when creating the `gradle` command args.


lib/index.js
```diff
@@ -111,10 +111,10 @@ function buildArgs(root, targetFile, gradleArgs, subProject) {
   args.push('-q');
   if (targetFile) {
     if (!fs.existsSync(path.resolve(root, targetFile))) {
-      throw new Error('File not found: ' + targetFile);
+      throw new Error('File not found: "' + targetFile + '"');
     }
     args.push('--build-file');
-    args.push(targetFile);
+    args.push('"' + targetFile + '"');
```
adrukh: This is the fix.

PR #26, opened 2019-03-04
fix:trimming off trailing (c) - resolution state

aarlaud:
- [x] Ready for review

#### What does this PR do?
trims off the trailing "(c)" that gradle dependencies command outputs in some cases, leading to missing out vulns.

run test/monitor a gradle project that contains (c) in one of it's dependency resolution. specially if it's a dependency with issues.
it will result on missing vulns on those dependencies. as we don't recognise the right package name.

### DIFF
--- a/lib/gradle-dep-parser.js
+++ b/lib/gradle-dep-parser.js
@@ -71,7 +71,8 @@ function processGradleOutput(text) {
 
       // trim (n) - Not resolved (configuration is not meant to be resolved)
       element = element.split(' (n)')[0];
-
+      // trim (c) - RESOLVED_CONSTRAINT
+      element = element.split(' (c)')[0];
       return element;
     });
   return {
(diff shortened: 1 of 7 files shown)


adrukh:
@aarlaud two things concern me here:
1. What is the meaning of `(c)` in the output we are parsing, and why is it safe to simply discard it?
2. Let’s add a test that will validate this important change so that we don’t run the risk of undoing it in the future.

aarlaud:
(c) means RESOLVED_CONSTRAINT. Details at https://docs.gradle.org/current/userguide/managing_transitive_dependencies.html and code at https://github.com/gradle/gradle/blob/811070ff1101e1b5070551713a0cef1d5409121f/subprojects/diagnostics/src/main/java/org/gradle/api/tasks/diagnostics/internal/graph/SimpleNodeRenderer.java#L29

It is probably useful information eventually but is currently tripping us up on the detection.

In terms of test, I'll need help setting something up.

PR #34, opened 2019-03-21
refactor: convert to Typescript (no proper lint, no types yet)

kyegupov:
- [ ] Ready for review

#### What does this PR do?

Converts the project to Typescript.
Most of the tslint rules disabled for now to keep the diff minimal.
No type definitions added yet.

We are working on modernizing Gradle support. To make sense of the logic we could use some type definitions.

Motivated by https://snyksec.atlassian.net/browse/BST-508

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -1,19 +1,17 @@
-var os = require('os');
-var fs = require('fs');
-var path = require('path');
-var subProcess = require('./sub-process');
+import * as os from 'os';
+import * as fs from 'fs';
+import * as  path from 'path';
+import * as subProcess from './sub-process';
+
 var packageFormatVersion = 'mvn:0.0.1';
 
-module.exports = {
-  inspect: inspect,
-};
-
-module.exports.__tests = {
+// tslint:disable-next-line
+export const __tests = {
   buildArgs: buildArgs,
   extractJsonFromScriptOutput: extractJsonFromScriptOutput,
 };
 
-function inspect(root, targetFile, options) {
+export function inspect(root, targetFile, options?) {
   if (!options) {
     options = {dev: false};
   }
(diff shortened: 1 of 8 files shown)


lib/index.ts
```diff
@@ -1,19 +1,17 @@
...
-var packageFormatVersion = 'mvn:0.0.1';
+import * as os from 'os';
+import * as fs from 'fs';
+import * as  path from 'path';
+import * as subProcess from './sub-process';
 
-module.exports = {
-  inspect: inspect,
-};
+var packageFormatVersion = 'mvn:0.0.1';
```
miiila: I wouldn't mind `var` to `const`, despite it wouldn't be pure refactoring then, but YMMV. (valid for the rest as well)

tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+  "compilerOptions": {
+      "outDir": "./dist",
+      "pretty": true,
+      "target": "es5",
+      "lib": [
+        "es5",
+        "es2015.promise",
+        "dom",
```
miiila: Do we need `dom` here? Do we manipulate it somewhere?

tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+      ],
+      "module": "commonjs",
+      "allowJs": false,
+      "sourceMap": true,
+      "strict": true,
```
miiila: We can add `declaration` to generate types, but we can also add it once we will really add some types :-)

miiila: It's also handy to add `importHelpers: true` and install `tslib` package; it makes compiled javascript smaller and more readable (if needed)

tsconfig.json
kyegupov: Nope. Mindlessly copied from a template. Droppin'.

lib/index.ts
kyegupov: Yeah, tslint should prohibit var, but enabling all the lints would pollute the diff, introduce a lot of work and little value. Better deferred to a later PR.

tsconfig.json
kyegupov: Hm, doesn't seem to make any change. But why not.

PR #37, opened 2019-03-22
[WIP] Feat/multiple projects

kyegupov:
- [ ] Ready for review


lili2311:
Please make sure all commits are named appropriately for release and squash to reduce the number of commits.

lib/index.ts
```diff
@@ -5,52 +5,128 @@ import * as subProcess from './sub-process';
...
 
-export interface SingleProjectResult {
+// TODO(kyegupov): the types below will be extracted to a common plugin interface library
+
+export interface BaseInspectOptions {
+  dev?: boolean;
+
+  // Additional command line arguments to Gradle, supplied after "--" to the Snyk CLI.
+  // E.g. --configuration=foo
+  args?: string[];
```
lili2311: Can we please rename this to something that won't require comments to explain what this is, for example `argsPassedDirectlyToGradle`?

Much prefer we rely on self explanatory naming instead of so many comments in the code.

lib/index.ts
```diff
@@ -5,52 +5,128 @@ import * as subProcess from './sub-process';
...
+}
+
+// The return type of inspect() depends on the multiDepRoots flag.
+
+export interface SingleRootInspectOptions extends BaseInspectOptions {
+  // Return the information not on the main project, but on the specific subproject defined in the build.
```
lili2311: unnessesary comment, this is self explanatory

lib/index.ts
```diff
@@ -5,52 +5,128 @@ import * as subProcess from './sub-process';
...
-  defaultProject: string;
-  projects: ProjectsDict;
+export interface DepDict {
+  [name: string]: DepTree;
 }
 
-interface ProjectsDict {
-  [project: string]: DepDict;
-}
+// TODO(darscan): proper name should be decided.
```
lili2311: Again this is too much commentary here that should live in a Jira ticket instead or somewhere else? I don't think this brings any value having possible new names listed here. As soon as we talk in person all these comments will become obsolete.

lib/index.ts
```diff
@@ -5,52 +5,128 @@ import * as subProcess from './sub-process';
...
-    },
-    package: pkg,
+  const plugin = {
+    name: 'bundled:gradle',
+    runtime: 'unknown',
+    targetFile: targetFileFilteredForCompatibility(targetFile),
   };
+  if ((options as MultiRootsInspectOptions).multiDepRoots) {
+    if (subProject) {
+      throw new Error('gradle-sub-project flag is incompatible with multiDepRoots');
```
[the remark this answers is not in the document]
kyegupov: Added a test (actually, Typescript also blocks that combination of flags).

lib/index.ts
```diff
@@ -5,52 +5,128 @@ import * as subProcess from './sub-process';
...
+// The return type of inspect() depends on the multiDepRoots flag.
+
+export interface SingleRootInspectOptions extends BaseInspectOptions {
+  // Return the information not on the main project, but on the specific subproject defined in the build.
+  'gradle-sub-project'?: string;
+}
+
+export interface MultiRootsInspectOptions extends BaseInspectOptions {
+  // Return multiple "dependency roots" as a MultiDepRootsResult.
+  // Dep roots correspond to sub-projects in Gradle or projects in a Yark workspace.
```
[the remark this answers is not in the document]
kyegupov: Imagine a person onboarded to the team. One of their starter tasks is to do something Gradle-related. They start to read the code to understand what the plugin API is. They are absolutely baffled by the combination of letters "multiDepRoots" (I would be!) and then go ahead to ask a fellow developer, distracting them and wasting time.

A few lines of comments in a critical place cost far less than developer time lost to confusion.

On top of that, it's a public repository, with possibly external contributors. They are in worse situation than even new starters.

In my life, I have never seen a project with too much documentation. A public interface, like this, should have tons of documentation.

lib/index.ts
```diff
@@ -5,52 +5,128 @@ import * as subProcess from './sub-process';
...
   };
+  if ((options as MultiRootsInspectOptions).multiDepRoots) {
+    if (subProject) {
+      throw new Error('gradle-sub-project flag is incompatible with multiDepRoots');
+    }
+    return {
+      plugin,
+      depRoots: await getAllDepsAllProjects(root, targetFile, options),
+    };
+  } else {
```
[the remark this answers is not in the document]
kyegupov: In general, I prefer aligned then-else blocks, but why not.

lib/index.ts
kyegupov: it's a public contract used by CLI, so - I don't think so.

lib/index.ts
kyegupov: If these comments become obsolete, we will write new ones.
But yes, Jira tickets would be better.

lib/index.ts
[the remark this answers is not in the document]
lili2311: Comments are not a blocker for release for me as long as you are happy to commit to keep them in sync and come back to amend as they become irrelevant

lib/index.ts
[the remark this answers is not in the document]
lili2311: 👍

lib/index.ts
lili2311: same as above as long as you are happy to keep them in sync

lili2311:
```Cannot read property 'depDict' of undefined``` => when trying to do `snyk test --gradle-sub-project=non-existent` we used to show the gradle build error, is it easy to align these to what it used to be the new error is too cryptic and doesn't tell the user what is wrong.

Ideally we want to say that that sub-project doesn't exists, not sure how easy that is

kyegupov:
@lili2311 added a nicer message for a missing subproj

kyegupov:
(rebased to latest)

PR #43, opened 2019-03-29
fix: support for Gradle versions before 4.3

kyegupov:
- [x] Ready for review

#### What does this PR do?

Fixes an issue where scans will fail because `--no-parallel` flag is not supported on Gradle <4.3

Run against projects with various versions of Gradle installed (natively and via wrapper)


lili2311:
Looks good, as per chat would be nice to have different version of gradle to run all the tests. But in a separate PR

PR #41, opened 2019-03-27
fix: properly support Snyk CLI standalone binaries

kyegupov:
- [x] Ready for review
- [x] Follows CONTRIBUTING rules
- [x] Reviewed by Snyk internal team

#### What does this PR do?

fix https://snyksec.atlassian.net/browse/BST-529

PR #46, opened 2019-03-29
fix: fix analysis in Windows standalone binaries

kyegupov:
- [x] Ready for review

#### What does this PR do?

Extension of the fix https://github.com/snyk/snyk-gradle-plugin/pull/41 for Windows.


lib/index.ts
```diff
@@ -88,9 +88,11 @@ async function getAllDeps(root, targetFile, options, subProject): Promise<DepTre
     throw new Error('Cannot locate Snyk init.gradle script');
   }
 
-  if (/^\/snapshot/.test(__dirname)) {
+  if (/^\/snapshot/.test(__dirname) || /^C:\\snapshot/.test(__dirname)) {
```
adrukh: This is quite brittle (UNIX vs Windows, implementation detail of how the virtual fs driver of pkg works, etc.)

Let's use a temp file *always* to be on the safe side? See https://github.com/snyk/snyk-python-plugin/pull/37 as a reference.

kyegupov: If you think that relying on temp file is a lower risk, let's do it, why not.

adrukh:
We’ve done so in our other plugins, please review the solution there and see if it makes you confident enough. Your hands on the wheel here :)

kyegupov:
@adrukh @lili2311 manually tested on Windows and Linux, appears to be working.

My worry is: what if this gets run in a read-only environment, with no access to temp filesystem? Do we conceivably have such use cases?

lib/index.ts
lili2311: Oh I did not know we did this in other plugins. good to know will reference them more. Agree with Anton here anything that makes this less risky / brittle is a much safer way forward and a nice experience for the user.

adrukh:
The concern is valid, but our experience with the other plugins gives me confidence that this will not be a degradation. Let’s release this one please.

PR #57, opened 2019-04-08
feat: for single-deproot scans, return also the names of all subprojects

kyegupov:
- [x] Ready for review

#### What does this PR do?
When doing a simple (one sub-project) scan, return the names of all subprojects, so that the CLI can advertise trying all-sub-project scans.

https://snyksec.atlassian.net/browse/BST-540

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -51,6 +51,13 @@ export interface PluginMetadata {
   // TODO(BST-542): remove, DepRoot.targetFile to be used instead
   // Note: can be missing, see targetFileFilteredForCompatibility
   targetFile?: string;
+
+  // Plugin-specific metadata
+  meta?: {
+    // If we don't return the results for all dependency roots (subprojects),
+    // still record their names to warn the user about them not being scanned
+    allDepRootNames?: string[];
+  };
 }
 
 export interface DepDict {
@@ -90,7 +97,7 @@ export async function inspect(root, targetFile, options?: SingleRootInspectOptio
   if (subProject) {
     subProject = subProject.trim();
   }
-  const plugin = {
+  const plugin: PluginMetadata = {
     name: 'bundled:gradle',
     runtime: 'unknown',
     targetFile: targetFileFilteredForCompatibility(targetFile),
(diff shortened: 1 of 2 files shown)


lib/index.ts
```diff
@@ -164,10 +168,13 @@ async function getAllDepsOneProject(root, targetFile, options, subProject): Prom
...
-    name: packageName,
-    version: packageVersion,
-    packageFormatVersion,
+    depTree: {
+      dependencies: depDict,
+      name: packageName,
+      version: packageVersion,
+      packageFormatVersion,
+    },
+    allDepRootNames: Object.keys(allProjectDeps.projects),
```
lili2311: could this live under meta? We agreed the new format is to have all meta like things under `meta`

lili2311:
What is the related PR in CLI please?  Can you link both so it is easier to see how it is used?

kyegupov:
@lili2311 https://github.com/snyk/snyk/pull/441

lib/index.ts
kyegupov: hmmm. Why not!

kyegupov:
Kicking the Travis build (status check points to a wrong run)

lib/index.ts
```diff
@@ -150,23 +158,31 @@ function extractJsonFromScriptOutput(stdoutText: string): JsonDepsScriptResult {
...
-    packageFormatVersion,
+    depTree: {
+      dependencies: depDict,
+      name: packageName,
+      // TODO: extract from project
+      // https://snyksec.atlassian.net/browse/BST-558
+      version: '0.0.0',
+      packageFormatVersion,
+    },
+    allDepRootNames,
```
lili2311: as per our new return type nothing should live on the root now. This should live in `meta` property.

lib/index.ts
```diff
@@ -51,6 +51,10 @@ export interface PluginMetadata {
   // TODO(BST-542): remove, DepRoot.targetFile to be used instead
   // Note: can be missing, see targetFileFilteredForCompatibility
   targetFile?: string;
+
+  // If we don't return the results for all dependency roots (subprojects),
+  // still record their names to warn the user about them not being scanned
+  allDepRootNames?: string[];
```
lili2311: please more to `meta`

lib/index.ts
kyegupov: Yes and it does. This function returns a pair of (deptree, alldeprootnames), and the latter is then stored in meta.

lib/index.ts
kyegupov: There is no `meta` field in the plugin response, there's only `plugin` and it's already in there.

lib/index.ts
lili2311: I was expecting to see something like `plugin.meta.allDepRootNames` not `plugin. allDepRootNames` because ideally all top level plugin return is the same across all languages and the meta holds unique properties.

kyegupov: hm, ok, let's try that for now

PR #59, opened 2019-04-09
fix: work around non-mergeable configs in Android builds

kyegupov:
- [x] Ready for review
- [x] Example fixture added (not automated tests)

#### What does this PR do?

* when specifying a concrete configuration, preserve its attributes
* if failing due to unmergeable configurations, display actionable suggestions for the user

index.ts and the message there; init.gradle; test/fixtures/android-cannot-auto-resolve/app/build.gradle 

run snyk test on various Gradle projects (android and not)

Should be all in the error message and comments.
If anything is not clear - let's add the comments.

https://snyksec.atlassian.net/browse/BST-528

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -273,14 +273,34 @@ async function getAllDeps(root, targetFile, options): Promise<JsonDepsScriptResu
       // intentionally empty
     }
     const orange = chalk.rgb(255, 128, 0);
+    const blackOnYellow = chalk.bgYellowBright.black;
     gradleVersionOutput = orange(gradleVersionOutput);
     const subProcessError = orange(error.message);
-    const mainErrorMessage = `Error running Gradle dependency analysis.
+    let mainErrorMessage = `Error running Gradle dependency analysis.
 
 Please ensure you are calling the \`snyk\` command with correct arguments.
 If the problem persists, contact support@snyk.io, providing the full error
 message from above, starting with ===== DEBUG INFORMATION START =====.`;
-    const blackOnYellow = chalk.bgYellowBright.black;
+
+    // Special case for Android, where merging the configurations is sometimes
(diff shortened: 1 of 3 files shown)


lib/index.ts
```diff
@@ -273,14 +273,34 @@ async function getAllDeps(root, targetFile, options): Promise<JsonDepsScriptResu
...
+
+It seems like you are scanning an Android build with ambiguous dependency variants.
+We cannot automatically resolve dependencies for such builds.
+
+We recommend converting your subproject dependency specifications from the form of
+    implementation project(":mymodule")
+to
+    implementation project(path: ':mymodule', configuration: 'default')
+or running Snyk CLI tool for a specific configuration, e.g.:
+    snyk test --all-sub-projects -- --configuration=releaseRuntimeClasspath`;
```
adrukh: Isn’t there a space instead of `=` before the configuration name?

kyegupov: @adrukh either way works.

adrukh: Very well!

kyegupov: (I'd rather this become a --configurationFilter=REGEXP flag for Snyk CLI instead of this undocumented pseudo-Gradle option, but I'd like to solve the immediate issue first)

adrukh: The idea was not to pull in every build tool option to become a snyk cli option, using the common standalone double-dash. Pros and cons for sure, please bring this up for a wider discussion on the team.

kyegupov: @adrukh `--configuration` is not a build tool option anymore, we do not pass it to Gradle directly but rather process it. It's time to start actually controlling it.

adrukh: If it’s ours, it’s ours :)

PR #64, opened 2019-04-29
Feat/test java 9

lili2311:
- [x] Ready for review

#### What does this PR do?
Test different gradle versions on java 9 too


test/system/multi-module.test.ts
```diff
@@ -86,12 +86,12 @@ test('multi-project: only sub-project has deps, none returned for main', async (
   t.notOk(result.package.dependencies);
 });
 
-test('multi-project: using gradle 3.0.0 via wrapper', async (t) => {
+test('multi-project: using gradle 5.4.1 via wrapper', async (t) => {
```
kyegupov: we don't need to emphasize here the gradle version anymore (we did before because we were concerned about gradle 3 in particular)

lili2311: updated

PR #67, opened 2019-05-09
fix: make the injected script scan only one sub-project

kyegupov:
- [x] Ready for review

#### What does this PR do?

Before, init.gradle script always scanned all subprojects. This means if one of the subprojects fails for some reason, all of them do.

Now, if it only needs to scan one project, it will scan one.

### DIFF
--- a/test/system/multi-module.test.ts
+++ b/test/system/multi-module.test.ts
@@ -65,7 +65,7 @@ test('multi-project: only sub-project has deps and they are returned', async (t)
     options);
   t.match(result.package.name, '/subproj',
     'sub project name is included in the root pkg name');
-  t.deepEqual(result.plugin.meta!.allDepRootNames, ['root-proj', 'subproj']);
+  t.deepEqual(result.plugin.meta!.allDepRootNames, ['subproj']);
 
   t.equal(result.package
     .dependencies!['com.android.tools.build:builder']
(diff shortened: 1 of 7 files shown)


test/system/multi-module.test.ts
```diff
@@ -65,7 +65,7 @@ test('multi-project: only sub-project has deps and they are returned', async (t)
     options);
   t.match(result.package.name, '/subproj',
     'sub project name is included in the root pkg name');
-  t.deepEqual(result.plugin.meta!.allDepRootNames, ['root-proj', 'subproj']);
```
lili2311: could we add a test that makes sure that:
- if a user has multiple sub-projects and all 3 have `compile` we can scan it OK with and without --all-sub-projects flag
- if a user has multiple sub-projects and only 1 has `compile` configuration then `--all-sub-projects` returns a correct error? What is expected behaviour in this scenario?

kyegupov: Actually I was wrong. Just a missing config is not a big deal, we already handle that. But there can be other error conditions: I will try to emulate them.

kyegupov: Added a test (but not for a missing configuration, for an unscannable sub-project).
In the other PR, I will try to replicate failure about conflicting build types for such a sub-project.

PR #71, opened 2019-05-23
feat: add debugging output to the injected script

kyegupov:
- [x] Ready for review

#### What does this PR do?
Enables realtime logging from within the `init.gradle` script.
Output lines prefixed by "ECHO " will be output as debug messages by the Snyk CLI.

init.gradle, especially comments at the top

`DEBUG=snyk-gradle-plugin snyk test` on a Gradle project.

Many of our customers experience Gradle analysis as pretty slow. We should investigate how the time is being spent.

image

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -5,6 +5,10 @@ import * as subProcess from './sub-process';
 import * as tmp from 'tmp';
 import {MissingSubProjectError} from './errors';
 import chalk from 'chalk';
+import debugModule = require('debug');
+
+// To enable debugging output, run the CLI as `DEBUG=snyk-gradle-plugin snyk ...`
+const debugLogging = debugModule('snyk-gradle-plugin');
 
 const packageFormatVersion = 'mvn:0.0.1';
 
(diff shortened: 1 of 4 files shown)


lib/index.ts
```diff
@@ -236,6 +240,15 @@ async function getAllDepsAllProjects(root, targetFile, options): Promise<DepRoot
   });
 }
 
+const reEcho = /^ECHO (.*)$/;
```
lili2311: maybe let's use the plugin name like we do for snyk? `snyk-gradle-plugin`

lib/index.ts
```diff
@@ -5,6 +5,10 @@ import * as subProcess from './sub-process';
 import * as tmp from 'tmp';
 import {MissingSubProjectError} from './errors';
 import chalk from 'chalk';
+import debugModule = require('debug');
+
+// To enable debugging output, run the CLI as `DEBUG=snyk-gradle-plugin snyk ...`
+const debugLogging = debugModule('snyk-gradle-plugin');
```
lili2311: we already have a debug mode, please print all this info if the user is passing `-d` to the CLI + plugin, let's not create more custom options as it is more effort for users to opt in and to configure their CIs etc.

kyegupov: Yes, it can be done, but the change will happen in the CLI when we bump the version of this plugin.

lib/index.ts
kyegupov: It can be no other plugin :) let's go with SNYKECHO, it should be clear and unambiguous enough

lib/index.ts
lili2311: Okay!

kyegupov: https://github.com/snyk/snyk/blob/master/src/cli/args.ts#L55

lili2311:
Please share an example output from this?

kyegupov:
@lili2311 Image added.

Also noted that we are showing this big error message from the plugin 4 times in the debug mode :-/ I'll also change the CLI to stop doing that.

PR #76, opened 2019-06-06
fix: print the Gradle command in debug output

kyegupov:
- [x] Ready for review

#### What does this PR do?
Prints the Gradle command to assist with debugging even the successful runs of CLI.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -203,6 +203,7 @@ function extractJsonFromScriptOutput(stdoutText: string): JsonDepsScriptResult {
   if (jsonLine === null) {
     throw new Error('No line prefixed with "JSONDEPS " was returned; full output:\n' + stdoutText);
   }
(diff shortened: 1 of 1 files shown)


lib/index.ts
```diff
@@ -203,6 +203,7 @@ function extractJsonFromScriptOutput(stdoutText: string): JsonDepsScriptResult {
   if (jsonLine === null) {
     throw new Error('No line prefixed with "JSONDEPS " was returned; full output:\n' + stdoutText);
   }
+  debugLog('The command produced JSONDEPS output of ' + jsonLine!.length + ' characters');
```
michael-go: is this the parsing of the output of the `gradle` command? this can be huge, no?

kyegupov: That's why I'm only printing the string length :) parsing is on the next line

michael-go: oh lol cool :)

PR #81, opened 2019-06-24
fix: scan all subprojects when selecting attributes for conifguration

kyegupov:
- [x] Ready for review

#### What does this PR do?
To make sure the constructed configuration is compatible with all the sub-projects of the project, we should scan all of them to find attributes that match user specification.

We even have a comment explaining why we should have been doing that, but we were only scanning a specified project. This has caused problems for some our customers.

Drive-by: recognise one more variant of Gradle error message related to unresolvable variants when displaying extended error message.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -26,6 +26,12 @@ const packageFormatVersion = 'mvn:0.0.1';
 const isWin = /^win/.test(os.platform());
 const quot = isWin ? '"' : '\'';
 
+const cannotResolveVariantMarkers = [
+  'Cannot choose between the following',
+  'Could not select value from candidates',
+  'Unable to find a matching variant of project',
+];
+
 // TODO(kyegupov): the types below will be extracted to a common plugin interface library
 
 export interface BaseInspectOptions {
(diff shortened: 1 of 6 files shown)


lib/init.gradle
```diff
@@ -154,7 +154,7 @@ allprojects { everyProj ->
                 // when resolving the project A, so that it selects a concrete variant of dependency B.
                 def allConfigurationAttributes = [:] // Map<Attribute<?>, Set<?>>
                 def attributesAsStrings = [:] // Map<String, Set<string>>
-                rootProject.allprojects.findAll(shouldScanProject).each { proj ->
+                rootProject.allprojects.each { proj ->
```
lili2311: test please :)

lib/index.ts
```diff
@@ -360,8 +366,7 @@ message from above, starting with ===== DEBUG INFORMATION START =====.`;
     // There are no automated tests for this yet (setting up Android SDK is quite problematic).
     // See test/manual/README.md
 
-    if (/Cannot choose between the following/.test(error.message)
-      || /Could not select value from candidates/.test(error.message)) {
+    if (cannotResolveVariantMarkers.find((m) => error.message.includes(m))) {
```
FauxFaux: Are these localised?

kyegupov: I haven't encountered localized Gradle yet. Is it a thing?

.gitignore
```diff
@@ -6,3 +6,9 @@ __pycache__/
 package-lock.json
 test/fixtures/**/build
 npm-debug.log
+
+# VS Code Java Plugin
```
lwywoo: Can these not go in your local gitignore config?

kyegupov: Yes, moved.

PR #82, opened 2019-06-28
BREAKING CHANGE: use the new plugin API, Typescript

kyegupov:
- [x] Ready for review

#### What does this PR do?
Uses "legacy" type definitions from https://github.com/snyk/snyk-cli-interface (some fields are renamed, so THIS WOULD ALSO REQUIRE CHANGES IN SNYK CLI).

Included:
* Typescript compilation is not only `strict`, but also `noImplicitAny`
* a "dry-run" tsc compilation of tests before running them
* delicious, home-baked type definitions for Tap

This paves the way for us to:
* better understand the interface between the plugins and the CLI
* improve the code quality of the plugins
* safely migrate to DepGraph without the fear of losing data.

https://github.com/snyk/snyk-cli-interface/blob/master/lib/legacy/plugin.ts
inspect function in index.ts in this repo

(CLI changes needed)

This paves the way for us to:
* better understand the interface between the plugins and the CLI
* improve the code quality of the plugins
* safely migrate to DepGraph without the fear of losing data.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -6,8 +6,13 @@ import * as tmp from 'tmp';
 import {MissingSubProjectError} from './errors';
 import chalk from 'chalk';
 import debugModule = require('debug');
+import {
+  SingleSubprojectInspectOptions, MultiSubprojectInspectOptions, SinglePackageResult, MultiProjectResult,
+  PluginMetadata, InspectOptions, InspectResult, isMultiSubProject,
+} from '@snyk/cli-interface/dist/legacy/plugin';
+import { DepTree, ScannedProject } from '@snyk/cli-interface/dist/legacy/common';
 
-// To enable debugging output, run the CLI as `DEBUG=snyk-gradle-plugin snyk ...`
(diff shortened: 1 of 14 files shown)


miiila:
@kyegupov Amazing! only comment for breaking change - it needs to be at the beginning of commit message body

lib/index.ts
```diff
@@ -6,8 +6,13 @@ import * as tmp from 'tmp';
 import {MissingSubProjectError} from './errors';
 import chalk from 'chalk';
 import debugModule = require('debug');
+import {
```
dkontorovskyy: This only thought for future. Maybe we need to change the way `@snyk/cli-interface` is doing export. Cos I would like not to see in my import path `dist` folder. So we need to have some `index.ts` that will export everything.

kyegupov: Yeah, I'm not a fan of that either. But re-exporting namespaces is not that easy in Typescript :(

I'm looking into the possible options.

PR #95, opened 2019-09-01
feat: get correct project name

### LINKED ISSUE #93: Extract gradle project & sub-project names on every test
### Expected behaviour
After doing a little digging realised we are not grabbing the actual project names when returning back results, here is the PR with some tests that shows the desired outcome https://github.com/snyk/snyk-gradle-plugin/tree/feat/get-correct-project-names

Users can scan a project with:
- `snyk test --all-sub-projects`
- `snyk test --file=path/to/build.sbt`
- `snyk test --gradle-sub-project=path/to/build.sbt`

So we need to be mindful of what is the context and grab the relevant project name that we just scanned.

There are two potentials:
- extract the project name in the already used `init.gradle`, there is most likely a way to get that there and return it back for processing in the TypeScript lib wrapper / child process wrapper
To get the correct project names for testing purposes you can run `gradle projects` which will list the root project and any sub-projects from that directory

- if that is not possible we can use `gradle` to get the names in the relevant context 🐌 this may be too slow, as we already execute gradle when getting the deps, so this would be a second call.

### Actual behaviour

The tests in this branch fail https://github.com/snyk/snyk-gradle-plugin/tree/feat/get-correct-project-names

### Steps to reproduce
Run the tests on the branch above

If applicable, please append the `--debug` flag on your command and include the output here **ensuring to remove any sensitive/personal details or tokens.

sfat:
- [x] Ready for review

What does this PR do?
Fixes #93 . Problem was the project name was not being picked correctly.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -5,8 +5,9 @@ import * as subProcess from './sub-process';
 import * as tmp from 'tmp';
 import {MissingSubProjectError} from './errors';
 import chalk from 'chalk';
+import {legacyCommon, legacyPlugin as api} from '@snyk/cli-interface';
 import debugModule = require('debug');
-import { legacyPlugin as api, legacyCommon } from '@snyk/cli-interface';
+
 type DepTree = legacyCommon.DepTree;
 type ScannedProject = legacyCommon.ScannedProject;
 
@@ -94,9 +95,11 @@ export async function inspect(
     if (subProject) {
       throw new Error('gradle-sub-project flag is incompatible with multiDepRoots');
     }
+    const scannedProjects = await getAllDepsAllProjects(root, targetFile, options);
+    plugin.meta = plugin.meta || {};
     return {
       plugin,
-      scannedProjects: await getAllDepsAllProjects(root, targetFile, options),
+      scannedProjects,
     };
   }
   const depTreeAndDepRootNames = await getAllDepsOneProject(root, targetFile, options, subProject);
@@ -107,6 +110,9 @@ export async function inspect(
   return {
     plugin,
     package: depTreeAndDepRootNames.depTree,
+    meta: {
+      gradleProjectName: depTreeAndDepRootNames.gradleProjectName,
+    },
   };
 }
 
@@ -153,20 +159,21 @@ function extractJsonFromScriptOutput(stdoutText: string): JsonDepsScriptResult {
 }
 
 async function getAllDepsOneProject(root: string, targetFile: string, options: Options, subProject?: string):
-    Promise<{depTree: DepTree, allSubProjectNames: string[]}> {
+    Promise<{depTree: DepTree, allSubProjectNames: string[], gradleProjectName: string}> {
   const packageName = path.basename(root);
   const allProjectDeps = await getAllDeps(root, targetFile, options);
   const allSubProjectNames = allProjectDeps.allSubProjectNames;
-  let depDict = {};
   if (subProject) {
+    const {depTree, meta} = getDepsSubProject(root, subProject, allProjectDeps);
     return {
-      depTree: getDepsSubProject(root, subProject, allProjectDeps),
+      depTree,
       allSubProjectNames,
+      gradleProjectName: meta.gradleProjectName,
     };
   }
 
-  depDict = allProjectDeps.projects[allProjectDeps.defaultProject].depDict;
-
+  const {projects, defaultProject} = allProjectDeps;
+  const {depDict} = projects[defaultProject];
   return {
     depTree: {
       dependencies: depDict,
@@ -177,26 +184,33 @@ async function getAllDepsOneProject(root: string, targetFile: string, options: O
       packageFormatVersion,
     },
     allSubProjectNames,
+    gradleProjectName: defaultProject,
   };
 }
 
-function getDepsSubProject(root: string, subProject: string, allProjectDeps: JsonDepsScriptResult) {
+function getDepsSubProject(root: string, subProject: string, allProjectDeps: JsonDepsScriptResult):
+    {depTree: DepTree, meta: any} {
   const packageName = `${path.basename(root)}/${subProject}`;
-  let depDict = {};
+  const gradleProjectName = `${allProjectDeps.defaultProject}/${subProject}`;
 
   if (!allProjectDeps.projects || !allProjectDeps.projects[subProject]) {
     throw new MissingSubProjectError(subProject, Object.keys(allProjectDeps));
   }
 
-  depDict = allProjectDeps.projects[subProject].depDict;
+  const depDict = allProjectDeps.projects[subProject].depDict;
 
   return {
-    dependencies: depDict,
-    name: packageName,
-    // TODO: extract from project
-    // https://snyksec.atlassian.net/browse/BST-558
-    version: '0.0.0',
-    packageFormatVersion,
+    depTree: {
+      dependencies: depDict,
+      name: packageName,
+      // TODO: extract from project
+      // https://snyksec.atlassian.net/browse/BST-558
+      version: '0.0.0',
+      packageFormatVersion,
+    },
+    meta: {
+      gradleProjectName,
+    },
   };
 }
 async function getAllDepsAllProjects(root: string, targetFile: string, options: Options): Promise<ScannedProject[]> {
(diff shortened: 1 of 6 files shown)


lib/index.ts
```diff
@@ -170,7 +169,7 @@ async function getAllDepsOneProject(root: string, targetFile: string, options: O
   return {
     depTree: {
       dependencies: depDict,
-      name: packageName,
+      name: allProjectDeps.defaultProject,
```
sfat: correct project name.

lili2311: is a similar change not needed on `getDepsSubProject`?

test/system/plugin.test.ts
```diff
@@ -98,7 +98,7 @@ test('tests for Gradle 3+', async (t0) => {
       const result = await inspect('.',
         path.join(fixtureDir('multi-config-attributes'), 'build.gradle'),
         {'configuration-attributes': 'usage:java-api'});
-      t.match(result.package.name, 'fixtures',
+      t.match(result.package.name, 'multi-config-attributes',
```
lili2311: what about the tests that are in this branch? https://github.com/snyk/snyk-gradle-plugin/compare/master...feat/get-correct-project-names?expand=1 Can you bring them over too to update every single test to expect correct project-name?

sfat: Well, I've created a branch from your branch branch that was already pushed with your changes, so it's only the diff that I've done and it's not showing up in this PR.
I can change to branch in which to merge to be master and that should show all the changes, both mine and yours.

lili2311: 👌 awesome do that so we can merge it all at once

sfat: updated 👍

.gitignore
```diff
@@ -7,3 +7,4 @@ __pycache__/
 package-lock.json
 test/fixtures/**/build
 npm-debug.log
+.idea/
```
kyegupov: Developer-specific ignores are best added to a global gitignore file:
https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore

sfat: Okay, so should I revert and just adding it on my local environment the global gitignore?
I was thinking that someone else will bump into the same issue and you'll end up commenting the same thing to that person 🙃 .
But if it's something that it's common that you are doing it's not a problem, I'll revert

lib/index.ts
sfat: I'll look into it, sure.

.gitignore
sfat: went ahead and removed the .idea and modified my local global .gitignore

kyegupov: Thank you!

lili2311:
@sfat before you merge or release new CLI, please could you verify the project works as expected by doing:
- `snyk monitor` using latest CLI version
- then again `snyk monitor` using the lasted CLI build locally (to build locally: `npm link` inside this repo & `npm link snyk-gradle-plugin` inside `snyk/snyk` repo.  And verify the same project gets updated and verify what changed in UI if anything visually.

sfat:
@lili2311 I was wondering how to do that.
I'll do that, thank you!

lib/index.ts
sfat: I was looking into `getDepsSubProject`, but I'm not sure what would be expected behaviour for the `packageName`.

Currently is `${path.basename(root)}/${subProject}`, where in this case for a sub project would be `./subproj`.

What I was thinking to instead to do change `packageName` to be `${allProjectDeps.defaultProject}/${subProject}`, which would result in `root-proj/subproj`.

Is this the expected behaviour?

kyegupov: @sfat Yes, this is probably the most sensible solution.

sfat: update addressed for `getDepsSubProject ` as well

sfat:
I've extended `DepTree` from `snyk-cli-interface` to include a new field called `newName` and made it an optional field.
`DepTree` seems a good place as it used both for single project, but also for multi project like projects. 
Let me know if I'm going in the right direction.

sfat:
> I've extended `DepTree` from `snyk-cli-interface` to include a new field called `newName` and made it an optional field.
> `DepTree` seems a good place as it used both for single project, but also for multi project like projects.
> Let me know if I'm going in the right direction.

I've talked yesterday with @miiila regarding the `DepTree` and he suggested to add the new name field in the `ScannedProject.meta`.

This will contain an array (maybe needed for multi level projects where you would have multiple names) of type map, where the key of the map is the old name and the value is the new name.

```
[
["multi-project-parallel","root-proj"],
["multi-project-parallel/subproj0","root-proj/subproj0"],
["multi-project-parallel/subproj1","root-proj/subproj1"],
["multi-project-parallel/subproj2","root-proj/subproj2"],
["multi-project-parallel/subproj3","root-proj/subproj3"],
["multi-project-parallel/subproj4","root-proj/subproj4"],
```

The reason for this change is that @miiila was saying that `DepTree` isn't an appropriate data structure for the old/new name convention and is more of a meta information that will be used elsewhere.

Again, let me know if this is okay or you have other ideas.

sfat:
related this this one: https://github.com/snyk/snyk-cli-interface/pull/20

lib/index.ts
```diff
@@ -153,20 +159,21 @@ function extractJsonFromScriptOutput(stdoutText: string): JsonDepsScriptResult {
 }
 
 async function getAllDepsOneProject(root: string, targetFile: string, options: Options, subProject?: string):
-    Promise<{depTree: DepTree, allSubProjectNames: string[]}> {
+    Promise<{depTree: DepTree, allSubProjectNames: string[], gradleProjectName: any}> {
```
lili2311: `gradleProjectName: string`

lib/index.ts
```diff
@@ -153,20 +159,21 @@ function extractJsonFromScriptOutput(stdoutText: string): JsonDepsScriptResult {
...
   const packageName = path.basename(root);
   const allProjectDeps = await getAllDeps(root, targetFile, options);
   const allSubProjectNames = allProjectDeps.allSubProjectNames;
-  let depDict = {};
   if (subProject) {
+    const depsSubProject = getDepsSubProject(root, subProject, allProjectDeps);
+    const depTree = depsSubProject.depTree;
```
lili2311: you can use destructuring here: `const {depTree} = depsSubProject` this will get prop `depTree` from `depsSubProject`

lib/index.ts
```diff
@@ -153,20 +159,21 @@ function extractJsonFromScriptOutput(stdoutText: string): JsonDepsScriptResult {
...
-      depTree: getDepsSubProject(root, subProject, allProjectDeps),
+      depTree,
       allSubProjectNames,
+      gradleProjectName: depsSubProject.meta.gradleProjectName,
     };
   }
 
-  depDict = allProjectDeps.projects[allProjectDeps.defaultProject].depDict;
-
+  const depDict = allProjectDeps.projects[allProjectDeps.defaultProject].depDict;
```
lili2311: same here, you can destructure which is prettier & preferred

lib/index.ts
```diff
@@ -153,20 +159,21 @@ function extractJsonFromScriptOutput(stdoutText: string): JsonDepsScriptResult {
...
 
 async function getAllDepsOneProject(root: string, targetFile: string, options: Options, subProject?: string):
-    Promise<{depTree: DepTree, allSubProjectNames: string[]}> {
+    Promise<{depTree: DepTree, allSubProjectNames: string[], gradleProjectName: any}> {
   const packageName = path.basename(root);
   const allProjectDeps = await getAllDeps(root, targetFile, options);
   const allSubProjectNames = allProjectDeps.allSubProjectNames;
-  let depDict = {};
   if (subProject) {
+    const depsSubProject = getDepsSubProject(root, subProject, allProjectDeps);
```
lili2311: actually you can now do `const {depTree, meta} = getDepsSubProject(root, subProject, allProjectDeps);` to get both properties back

lib/index.ts
```diff
@@ -205,8 +219,13 @@ async function getAllDepsAllProjects(root: string, targetFile: string, options:
   const packageVersion = '0.0.0';
   return Object.keys(allProjectDeps.projects).map((proj) => {
     const packageName = proj === allProjectDeps.defaultProject ? basePackageName : basePackageName + '/' + proj;
+    const defaultProject = allProjectDeps.defaultProject;
+    const gradleProjectName = proj === defaultProject ? defaultProject : defaultProject + '/' + proj;
```
lili2311: You can use a string literal here like this: ``{$defaultProject}/${proj}`

test/system/multi-module.test.ts
```diff
@@ -202,8 +216,10 @@ test('multi-project: parallel with allSubProjects produces multiple results with
   const result = await inspect(fixtureDir('multi-project-parallel'), 'build.gradle', {allSubProjects: true});
   t.equal(result.scannedProjects.length, 6);
   const names = new Set<string>();
+  const newNames = new Set<string>();
```
lili2311: just curious why set?

sfat: just took the example for how `names` was defined, no reason in particular

lib/index.ts
sfat: good idea! will change it

lib/index.ts
sfat: sure thing

test/system/multi-module.test.ts
kyegupov: So that we don't care about the order of the elements in the set when we are comparing it.

The order of "scannedProjects" depends on the order of iteration of Object.keys(), which is not deterministic.

Issue #89, opened 2019-08-07
Store gradle & java version meta on every plugin execution


lili2311:
### Add more meta on every test run
In order to have better `-d` output in the CLI for support & to have a general idea of popular version of java & gradle that should be supported it would be good to save this data as meta on the plugin response.

Update the plugin to:

- [ ] run the `gradle -v command` in the plugin before or after each test run
- [ ] parse the output of it and set it as meta to be returned back with results
- [ ] add a test ito verify parsing is working as expected
- [ ] add tests to assert this meta is being returned by the plugin
- [ ] Snyk team to add the changes in the app UI to render the new meta on the project page

`gradle -v`
```
Gradle 4.10.2

Build time:   2018-09-19 18:10:15 UTCRevision:     b4d8d5d170bb4ba516e88d7fe5647e2323d791dd

Kotlin DSL:   1.0-rc-6Kotlin:       1.2.61
Groovy:     2.4.15
Ant:          Apache Ant(TM) version 1.9.11 compiled on March 23 2018
JVM:         10.0.1 ("Oracle Corporation" 10.0.1+10)
OS:           Mac OS X 10.13.2 x86_64
```

Returned Meta  should match:
```
const pluginMeta = {
  meta: {
    gradle: '4.10.2',
    jvm: '10.0.1',
    os: 'Mac OS X 10.13.2 x86_64',
    groovy: '2.4.15',
  }
}
```
Please also consider `Kotlin`, so instead of `groovy ` prop we can expect `kotlin`?

adrukh:
Please consider the case where a tested project carries a gradle wrapper `gradlew`, setting a specific version of gradle for the project.

sfat:
@lili2311 
I've messed around and parsed all the `gradle -v` output:
I've created an interface containing this information:
```
{
   "gradleVersion":"4.10.3",
   "metaBuildVersion":[
      {
         "meta":"Build time",
         "value":"2018-12-05 00:50:54 UTC"
      },
      {
         "meta":"Revision",
         "value":"e76905e3a1034e6f724566aeb985621347ff43bc"
      },
      {
         "meta":"Kotlin DSL",
         "value":"1.0-rc-6"
      },
      {
         "meta":"Kotlin",
         "value":"1.2.61"
      },
      {
         "meta":"Groovy",
         "value":"2.4.15"
      },
      {
         "meta":"Ant",
         "value":"Apache Ant(TM) version 1.9.11 compiled on March 23 2018"
[...]

PR #96, opened 2019-09-21
feat: store gradle version output meta info

sfat:
- [ ] Ready for review

#### What does this PR do?
implementation for https://github.com/snyk/snyk-gradle-plugin/issues/89

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -5,8 +5,9 @@ import * as subProcess from './sub-process';
 import * as tmp from 'tmp';
 import {MissingSubProjectError} from './errors';
 import chalk from 'chalk';
+import {legacyCommon, legacyPlugin as api} from '@snyk/cli-interface';
 import debugModule = require('debug');
-import { legacyPlugin as api, legacyCommon } from '@snyk/cli-interface';
+
 type DepTree = legacyCommon.DepTree;
 type ScannedProject = legacyCommon.ScannedProject;
 
@@ -94,9 +95,11 @@ export async function inspect(
     if (subProject) {
       throw new Error('gradle-sub-project flag is incompatible with multiDepRoots');
     }
+    const scannedProjects = await getAllDepsAllProjects(root, targetFile, options);
+    plugin.meta = plugin.meta || {};
     return {
       plugin,
-      scannedProjects: await getAllDepsAllProjects(root, targetFile, options),
+      scannedProjects,
     };
   }
   const depTreeAndDepRootNames = await getAllDepsOneProject(root, targetFile, options, subProject);
@@ -107,6 +110,9 @@ export async function inspect(
   return {
     plugin,
     package: depTreeAndDepRootNames.depTree,
+    meta: {
+      gradleProjectName: depTreeAndDepRootNames.gradleProjectName,
+    },
   };
 }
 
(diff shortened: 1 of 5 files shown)


sfat:
linked `snyk-gradle-plugin` with my local `snyk`
Added this snippet in `run-test.ts` to check that `snyk test -d` spits the needed information:

```
    if (deps.plugin && deps.plugin.meta && deps.plugin.meta.versionBuildInfo) {
      analytics.add('gradleVersion', deps.plugin.meta.versionBuildInfo.gradleVersion);
      const metaBuildVersion = deps.plugin.meta.versionBuildInfo.metaBuildVersion;
      if (metaBuildVersion) {
        analytics.add('metaBuildVersion', metaBuildVersion);
      }
    }
```
Screenshot with the output from `snyk test -d`
Screenshot 2019-09-20 at 21 46 48

sfat:
the jobs fails because this needs review first: https://github.com/snyk/snyk-cli-interface/pull/21

sfat:
Changed implementation to saved in `MultiProjectResult.meta` instead of `plugin.meta`.
Screenshot 2019-10-21 at 12 17 12

To accomplish the above screenshot, I've modified `src/lib/snyk-test/run-test.ts`, in `assembleLocalPayloads()` the following:

```
    for (const scannedProject of deps.scannedProjects) {
      if (scannedProject && scannedProject.meta && scannedProject.meta.buildInfoVersion) { // from here
        analytics.add('gradleVersion', scannedProject.meta.buildInfoVersion.gradleVersion);
        const metaBuildVersion = scannedProject.meta.buildInfoVersion.metaBuildVersion;
        if (metaBuildVersion) {
          analytics.add('metaBuildVersion', JSON.stringify(metaBuildVersion));
        }
      }
```

lib/index.ts
```diff
@@ -225,6 +235,7 @@ async function getAllDepsAllProjects(root: string, targetFile: string, options:
       targetFile: targetFileFilteredForCompatibility(allProjectDeps.projects[proj].targetFile),
       meta: {
         gradleProjectName,
+        buildInfoVersion: allProjectDeps.versionBuildInfo,
```
lili2311: what is the difference between `versionBuildInfo` and `buildInfoVersion`?

lib/index.ts
```diff
@@ -304,7 +348,9 @@ async function getAllDeps(root: string, targetFile: string, options: Options):
     if (cleanupCallback) {
       cleanupCallback();
     }
-    return extractJsonFromScriptOutput(stdoutText);
+    const extractedJson = extractJsonFromScriptOutput(stdoutText);
+    extractedJson.versionBuildInfo = getVersionBuildInfo(gradleVersionOutput);
```
lili2311: what is the parsing of build info fails? Will the whole test fail? Can we make it fail silently but log the failure?

lib/index.ts
sfat: yeah, I wasn't decided which name would be more appropriated and I think now there are 2 flavours.
Any preference?

lib/index.ts
sfat: yeah, I can make it fail silently.

lib/index.ts
```diff
@@ -382,6 +428,13 @@ ${chalk.red.bold(mainErrorMessage)}`;
   }
 }
 
+function toCamelCase(input: string) {
+  input = input.toLowerCase().replace(/(?:(^.)|([-_\s]+.))/g, (match: string) => {
+    return match.charAt(match.length - 1).toUpperCase();
+  });
+  return input.charAt(0).toLowerCase() + input.substring(1);
```
lili2311: is this safe? `.toLowerCase()` will fail on null/bad input

lib/index.ts
sfat: Since I've named it in `cli-interface` `versionBuildInfo`, I'll leave the same everywhere.
Btw, changed.

lib/index.ts
sfat: I tested with empty string and it behaving okay. 
Is there a workaround to test with something like `undefined`? I've tried adding it in my tests, but it will not compiling complaining it cannot apply `undefined` for type `string`

lib/index.ts
sfat: for the `fail silently` bit, do I just use `debugLog()` or something else?

lili2311: exactly add a `debugLog` and keep going so we can find this if it is run  with `-d`

lib/index.ts
lili2311: if you are confident then that is good enough :)

lib/index.ts
sfat: wrapping `getVersionBuildInfo` in a try catch block and just doing a `debugLog` to avoid failing of `test`.

lib/index.ts
```diff
@@ -304,7 +352,13 @@ async function getAllDeps(root: string, targetFile: string, options: Options):
     if (cleanupCallback) {
       cleanupCallback();
     }
-    return extractJsonFromScriptOutput(stdoutText);
+    const extractedJson = extractJsonFromScriptOutput(stdoutText);
+    try {
+      extractedJson.versionBuildInfo = getVersionBuildInfo(gradleVersionOutput);
+    } catch (error) {
+      debugLog('version build info not present, skipping ahead: ' + error);
```
lili2311: 👍

PR #103, opened 2019-11-06
fix: fix gradle for jdk8

sfat:
- [x] Ready for review

#### What does this PR do?
JDK8 is not being installed before being used.
Also, since SDKMAN will change versions and removing old ones, it make sense to make the installation of the version a bit more generic

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ before_script:
   - echo sdkman_auto_answer=true > ~/.sdkman/etc/config
   # install jdk is needed for JDK9+
   - wget https://github.com/sormuras/bach/raw/master/install-jdk.sh
+  - unset JAVA_HOME
 jobs:
   include:
   - stage: Node versions + actual Gradle version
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -53,22 +53,22 @@ jobs:
     # java 9 and gradle 2 not supported
     env: GRADLE_VERSION=2.14 JDK=8
     script:
-      - sdk use java 8.0.232.hs-adpt
+      - sdk install java $(sdk list java | grep -o "$JDK\.[0-9\.]*hs-adpt" | head -1)
```
FauxFaux: `grep -m1` is useful for this.

sfat: changed!

.travis.yml
```diff
@@ -53,22 +53,22 @@ jobs:
     # java 9 and gradle 2 not supported
     env: GRADLE_VERSION=2.14 JDK=8
     script:
-      - sdk use java 8.0.232.hs-adpt
+      - sdk install java $(sdk list java | grep -o -m1 "$JDK\.[0-9\.]*hs-adpt")
```
lili2311: so what version will it pick here?

sfat: at the moment, it will pick `8.0.232.hs-adpt` as before.
But the way SDKMAN works, it seems, is that when a new version of JDK 8 will appear, the current version will be deleted, hence, we'll have to modify this travis file again.
the `$JDK` in this case is set to 8. We are making use of the `env` variable set at line 55.

PR #106, opened 2020-01-07
fix: plain console mode

pavel-github:
- [x] Ready for review

#### What does this PR do?
This PR forces `plain` console mode when executing Gradle commands. If it's not set, then build output contains control ANSI characters and generated JSON cannot be parsed.

This PR was triggered by failed issues in teamcity plugin.

Compare 2 Travis builds:
- failed build after console mode is manually set from outside: https://travis-ci.org/snyk/snyk-gradle-plugin/builds/633516869
- passed build after console mode is set for each Gradle command executing https://travis-ci.org/snyk/snyk-gradle-plugin/builds/633540774

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ before_script:
   - source "$HOME/.sdkman/bin/sdkman-init.sh"
   - echo sdkman_auto_answer=true > ~/.sdkman/etc/config
   - unset JAVA_HOME
+  - export GRADLE_OPTS=-Dorg.gradle.console=verbose
 jobs:
   include:
   - stage: Different Node/Java versions + most recent Gradle version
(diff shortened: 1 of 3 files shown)


lili2311:
@pavel-github could you please also attach a screenshot of the difference you get today when running with CLI before and with this change? Will the user see anything different?

pavel-github:
@lili2311, I have only screenshot with "fail-behavior" because our teamcity plugin has bundled CLI.
teamcity__build-output-with-verbose-console

lili2311:
@pavel-github I was specifically wondering how does this affect the cli, you can test it by locally linking this branch with `npm link` and then running `npm link snyk-gradle-plugin` in the locally built cli and test the difference between the console of released cli and this

gitphill:
paired with @pavel-github to show how to generate the screen shots you're after @lili2311 
@pavel-github looks like we need to rebase against master here too to resolve .travis.yaml merge conflict

pavel-github:
@gitphill, yes. I fixed a typo on 'distribution' word :).

pavel-github:
"bad" case:
snyk-gradle_without-fix

good case:
snyk-gradle_with-fix

.travis.yml
```diff
@@ -18,15 +18,17 @@ before_script:
   - source "$HOME/.sdkman/bin/sdkman-init.sh"
   - echo sdkman_auto_answer=true > ~/.sdkman/etc/config
   - unset JAVA_HOME
+  - export GRADLE_OPTS=-Dorg.gradle.console=verbose
```
gitphill: we don't need this since we override the env var?

gitphill: or is this to test that we do override?

pavel-github: yes, if we don't override, then the verbose console mode will break builds.

PR #109, opened 2020-01-30
Update init.gradle

RemonSinnema:
Processed dependencies can be found later as transitive dependencies of other dependencies.

- [ ] Ready for review

#### What does this PR do?

Fix issue with OOM when there are many dependencies with the same transitive dependencies, e.g. using Spring Boot.

https://support.snyk.io/hc/en-us/requests/3355

### DIFF
--- a/lib/init.gradle
+++ b/lib/init.gradle
@@ -82,7 +82,6 @@ allprojects { everyProj ->
                     def row = ['name': depName, 'version': d.moduleVersion, 'conf': d.configuration]
                     currentChain.add(depName)
                     def subDeps = depsToDict(d.children, currentChain)
-                    currentChain.remove(depName)
                     if (subDeps.size() > 0) {
                         row['dependencies'] = subDeps
                     }


lili2311:
👋 @RemonSinnema thanks for this! Would you be able to also add a test that shows the issue and that it is now fixed? Or at least share an example project that is buildable to verify before & after

RemonSinnema:
See the support request https://support.snyk.io/hc/en-us/requests/3355. Without the fix, I get an OOM. With the fix I don't. Unfortunately, I can't share the project I'm testing this on...

lili2311:
Great! no worries, any chance of a slimmed down test fixture instead?

lili2311:
Thank you! closing this PR and used your commit in linked PR to get the tests to run

lili2311:
@RemonSinnema closing this PR as this breaks existing functionality, let's keep chatting on the support case to fins a solution.

anthogez:
@RemonSinnema We are current working on Gradle CLI Improvements

PR #113, opened 2020-02-25
Feat/dot notation

aviadhahami:
With very large projects, this initscript has some problems with regards
to memory usage since it is attempting to keep the entire dependency
tree in memory through any included configurations or attributes. In my
Gradle version at least it also appeared that the snykMergedDepsConf
would result in a cycle, but dependency cycles in Gradle are not
allowed within a configuration, so the combination of all configurations
should also not have any cycles.

- [ ] Ready for review

#### What does this PR do?
Converts the graph generation to use dot format. This alleviates the memory consumption problem for large projects as well as speeding up graph generation. 

Most of the code is in the init.gradle file. However, the parse-gradle.ts was scavenged pretty close to verbatim from the [snyk-mvn-plugin](https://github.com/snyk/snyk-mvn-plugin).

I wasn't sure how to run the plugin through the CLI directly, so I was mostly running this directly via the Gradle commands: 

```
./gradlew -I lib/init.gradle snykResolvedDeps -q --no-daemon -Dorg.gradle.parallel=
./gradlew -I lib/init.gradle snykResolvedDeps -q --no-daemon -Dorg.gradle.parallel= -Pconfiguration=testRuntime
./gradlew -I lib/init.gradle snykResolvedDeps -q --no-daemon -Dorg.gradle.parallel= -PonlySubProject=sub-project
```

All commands should remain identical to their previous counterparts.

Snyk support ticket 3464


lib/index.ts
```diff
@@ -159,10 +155,10 @@ function extractJsonFromScriptOutput(stdoutText: string): JsonDepsScriptResult {
       jsonLine = l.substr(9);
     }
   });
-  if (jsonLine === null) {
+  if (!jsonLine === null) {
```
gitphill: intentional?

shanman190:
@aviadhahami and @orsagie, if there's anything I can do to help please let me know. I'm more than eager to learn! :)

lib/index.ts
aviadhahami: image
fixed

lib/parse-gradle-new.ts
```diff
@@ -0,0 +1,83 @@
...
+const LOG_LABEL = /^SNYKDEPS\s*/gm;
+const DIGRAPH_SEGMENT = /digraph\W([\s\S]*?)/;
+const GRADLE_DEP = /([^:]+):([^:]+):([^:]+):([^:]+)(?:([^:]+))?/m;
+const ARROW_SEPARATOR = '->';
+
+function depTreeFromGradleString(s: string): DepTreeDep {
+    const pkgMeta: RegExpMatchArray | string[] = s.match(GRADLE_DEP) || [];
+
+    // 0 matches
+    // 1 -> package name
```
shanman190: 1 -> maven group
2 -> maven artifact id
3 -> packaging type (most often will be jar, but there are a few other variants)
4 -> version
5 -> configuration for artifact

shanman190: Just an FYI, from a Snyk standpoint #5 (configuration for artifact) is always going to come out as runtime because the plugin is extending the Gradle `runtime` configuration with it's custom `snykConf` configuration. Since Snyk, for now doesn't care about which Java configuration that the dependency comes from, this should be ok to pretty well ignore.

anthogez:
Hi @shanman190 ,
I would like to update you that we are current working on Gradle Improvements :)

shanman190:
I've also got availability to help if anything is needed. :)

anthogez:
@shanman190 I would like to say thank you very much for your insights and collaboration with our team, but we decided of move forward with an alternative solution that is current available on the latest version of snyk-cli, would be nice to have some feedback :)

Thanks again!

shanman190:
Yep. I've been keeping up with it. :)

The first time I tested I had raised the issue on the Gradle wrapper bug. The model from a Gradle standpoint was much more performance though at that time it was still using the depTree format so was very nodejs memory intensive. I'll try out the depGraph updates tomorrow and let you know how that testing goes for me.

I do think the PRs that you've been putting in supercede the work done here quite nicely and as always more than happy to jump in and get my handy dirty! :)

anthogez:
> Yep. I've been keeping up with it. :)
> 
> The first time I tested I had raised the issue on the Gradle wrapper bug. The model from a Gradle standpoint was much more performance though at that time it was still using the depTree format so was very nodejs memory intensive. I'll try out the depGraph updates tomorrow and let you know how that testing goes for me.
> 
> I do think the PRs that you've been putting in supercede the work done here quite nicely and as always more than happy to jump in and get my handy dirty! :)

I agree the usage of gradle wrapper is important, I am happy to hear it thanks!
I am looking forward for your feedback :)

PR #119, opened 2020-05-18
fix: gradle oom issue

anthogez:
- [x] Ready for review

#### What does this PR do?
This PR, brings the initial support of graphs to this snyk-cli plugin.
At the moment with the change of data structure from Tree to Graph on the gradle task, we are avoiding OOM on java heap while executing it to load a huge amount of projects, but while we do not change snyk-cli contract rules to allow it of receives Acyclic depGraphs from plugins, we are converting processed graphs back to depTree out of the gradle task. Once the contract rules of cli changes, the  convertDepGraphToDepTree method will be deleted.

I would like to point out that at the moment only Gradle v3+ will be using the initial new graph support, we kept for now v2 with the previously gradle task that we had and it's now considered as legacy.

With gradual steps, tests will be rewritten more concise with Jest and with new purposes too.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -5,6 +5,7 @@ import * as subProcess from './sub-process';
 import * as tmp from 'tmp';
 import { MissingSubProjectError } from './errors';
 import * as chalk from 'chalk';
+import { DepGraphBuilder, PkgManager, DepGraph, legacy } from '@snyk/dep-graph';
 import { legacyCommon, legacyPlugin as api } from '@snyk/cli-interface';
 import debugModule = require('debug');
 
@@ -191,6 +192,67 @@ function extractJsonFromScriptOutput(stdoutText: string): JsonDepsScriptResult {
   return JSON.parse(jsonLine!);
 }
 
+async function depsFromGraph(graphData: any) {
+  if (!graphData || Object.keys(graphData).length === 0) {
+    return {};
+  }
+
+  const pkgManager: PkgManager = { name: 'gradle' };
+  const builder = new DepGraphBuilder(pkgManager, {
+    name: 'root',
+    version: '1.0.0',
+  });
+
+  for (const key of Object.keys(graphData)) {
+    const { name, version } = graphData[key];
+    const nodeId = name;
+    // adding nodes
+    builder.addPkgNode(
+      {
+        name,
+        version,
+      },
+      nodeId,
+    );
+  }
+
+  for (const key of Object.keys(graphData)) {
+    const { name, parentNames } = graphData[key];
+    const nodeId = name;
+    // adding edges
+    if (parentNames && parentNames.length > 0) {
+      for (const parentName of parentNames) {
+        builder.connectDep(parentName, nodeId);
+      }
+    } else {
+      builder.connectDep('root-node', nodeId);
+    }
+  }
+
+  const depGraph = builder.build();
+  const depTree: legacy.DepTree = await convertDepGraphToDepTree(
+    depGraph,
+    'gradle',
+  );
+  return depTree.dependencies as any;
+}
+
+export async function convertDepGraphToDepTree(
+  depGraph: DepGraph,
+  type: legacyCommon.SupportedPackageManagers,
+  pruneGraph = false,
+): Promise<legacy.DepTree> {
+  try {
+    const depTree = (await legacy.graphToDepTree(depGraph, type, {
+      deduplicateWithinTopLevelDeps: pruneGraph,
+    })) as DepTree;
+    return depTree;
+  } catch (e) {
+    const errorMessage = 'Failed converting depGraph to depTree';
+    throw new Error(errorMessage);
+  }
+}
+
 async function getAllDepsOneProject(
   root: string,
   targetFile: string,
@@ -317,14 +379,24 @@ async function getInjectedScriptPath(): Promise<{
   cleanupCallback?: () => void;
 }> {
   let initGradleAsset: string | null = null;
+  const gradleVersionOutput = await subProcess.execute('gradle', ['-v'], {});
+  const isGradle3Plus =
+    parseInt(gradleVersionOutput.match(/Gradle (\d+)\.\d+(\.\d+)?/)![1], 10) >=
+    3;
   if (/index.js$/.test(__filename)) {
     // running from ./dist
     // path.join call has to be exactly in this format, needed by "pkg" to build a standalone Snyk CLI binary:
     // https://www.npmjs.com/package/pkg#detecting-assets-in-source-code
     initGradleAsset = path.join(__dirname, '../lib/init.gradle');
+    if (!isGradle3Plus) {
+      initGradleAsset = path.join(__dirname, '../lib/legacy-init.gradle');
+    }
   } else if (/index.ts$/.test(__filename)) {
     // running from ./lib
     initGradleAsset = path.join(__dirname, 'init.gradle');
+    if (!isGradle3Plus) {
+      initGradleAsset = path.join(__dirname, 'legacy-init.gradle');
+    }
   } else {
     throw new Error('Cannot locate Snyk init.gradle script');
   }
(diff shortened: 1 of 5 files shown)


test/system/multi-module.test.ts
```diff
@@ -185,6 +193,19 @@ test('multi-project: only sub-project has deps and they are returned space needs
     options,
   );
 
+  console.log(
+    'WHATS WRONG' +
```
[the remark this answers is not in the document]
anthogez: No worries I knew what I was logging, I just sent it by mistake here but already deleted :)
Just wrote random phrase while I was checking the results ahaha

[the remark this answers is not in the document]
ekbsnyk: 👍 wondered!

lib/index.ts
```diff
@@ -191,6 +191,26 @@ function extractJsonFromScriptOutput(stdoutText: string): JsonDepsScriptResult {
   return JSON.parse(jsonLine!);
 }
 
+function depsToDict(obj: any) {
+  const res: any = {};
```
gitphill, line 195: can we type this?

anthogez: Sure, was taking care of tests first :)

lib/index.ts
```diff
@@ -191,6 +197,74 @@ function extractJsonFromScriptOutput(stdoutText: string): JsonDepsScriptResult {
...
+  const depGraph = builder.build();
+  const depTree: legacy.DepTree = await convertDepGraphToDepTree(
+    depGraph,
+    'gradle',
+  );
+
+  return depTree.dependencies as any;
+}
+
+export async function convertDepGraphToDepTree(
```
anthogez, line 251: This was used to see if graph values were correct

lib/index.ts
```diff
@@ -191,6 +192,67 @@ function extractJsonFromScriptOutput(stdoutText: string): JsonDepsScriptResult {
...
+
+  const depGraph = builder.build();
+  const depTree: legacy.DepTree = await convertDepGraphToDepTree(
+    depGraph,
+    'gradle',
+  );
+  return depTree.dependencies as any;
+}
+
+export async function convertDepGraphToDepTree(
```
anthogez, line 240: This will be the first approach, once Snyk-CLI changes of contract allows the direct usage of graphs, this convertDepGraph to DepTree will be removed

anthogez, line 240: This will be the first approach, once we apply contract changes on Snyk CLI, we will be allowed of send depGraphs from plugins to cli, and this convertDepGraph to DepTree will be removed

anthogez, line 240: This will be the first approach, once we apply contract changes on Snyk CLI, we will be allowed of send depGraphs from plugins to cli, and this convertDepGraph to DepTree will be deleted.

lib/legacy-init.gradle
```diff
@@ -0,0 +1,239 @@
+import groovy.json.JsonOutput
```
anthogez, line 1: This is the previously init.gradle that was replaced with the new one, but we are keeping it to keep the support  of gradle v2

lib/index.ts
```diff
@@ -191,6 +192,67 @@ function extractJsonFromScriptOutput(stdoutText: string): JsonDepsScriptResult {
   return JSON.parse(jsonLine!);
 }
 
+async function depsFromGraph(obj: any) {
```
anthogez, line 195: this `any type` will be replaced in a side pr, does not make sense pass the current available type, it uses a depTree while the structure that arrives on this obj is of a graph generated on gradle task.

orsagie, line 195: Can you change the naming in this method to be more descriptive? like obj -> graph and key -> package etc

lib/index.ts
```diff
@@ -191,6 +192,67 @@ function extractJsonFromScriptOutput(stdoutText: string): JsonDepsScriptResult {
...
+      builder.connectDep('root-node', nodeId);
+    }
+  }
+
+  const depGraph = builder.build();
+  const depTree: legacy.DepTree = await convertDepGraphToDepTree(
+    depGraph,
+    'gradle',
+  );
+  return depTree.dependencies as any;
```
orsagie, line 237: why are we only returning the dependencies?

lib/index.ts
```diff
@@ -438,6 +510,17 @@ async function getAllDeps(
     if (versionBuildInfo) {
       extractedJson.versionBuildInfo = versionBuildInfo;
     }
+    const { gradleVersion } = extractedJson?.versionBuildInfo;
+    // gradleVersion can be e.g. 6.4.1, below we are taking first number and validating.
+    const isGradle3Plus = gradleVersion && +gradleVersion?.substring(0, 1) >= 3;
+    // only gradle v3+ will begin with graph support while scanning projects with gradle task
+    if (isGradle3Plus) {
+      for (const key in extractedJson.projects) {
```
orsagie, line 518: key -> projectName/project/ whatever it actually is please

lib/init.gradle
```diff
@@ -54,6 +54,72 @@ import java.util.regex.Matcher
...
+       if(!childNode) {
+           return
+       }
+       childNode.parentKeys.push(parentKey)
+   }
+}
+
+def loadGraph(Iterable deps, SnykGraph graph, parentKey) {
+        deps.each { d -> 
+            def childKey = "${d.moduleGroup}:${d.moduleName}"
```
orsagie, line 105: if key is packageName, can we change to that all over?

lib/index.ts
anthogez: because this is the exactly data that we need to assign as result to keep the consistency of the expected result, with the previously gradle task we were just returning a tree so it's ok for this case (we are just returning the tree of deps)

test/system/multi-module.test.ts
```diff
@@ -152,7 +163,9 @@ if (wrapperIsCompatibleWithJvm) {
       'root-proj',
       'subproj',
     ]);
-    t.notOk(result.package.dependencies);
+    if (result.package.dependencies) {
```
orsagie, line 166: doesn't this defeat the purpose of the test? If this is for specific gradle version, can we change the check to be based on that

test/system/multi-module.test.ts
```diff
@@ -415,26 +443,28 @@ test('multi-project-dependency-cycle: scanning the main project works fine', asy
...
 
-  t.equal(
-    result.package.dependencies!['com.github.jitpack:subproj'].dependencies![
-      'com.github.jitpack:root-proj'
-    ].version,
-    'unspecified',
-    'dependency cycle is returned in the results',
-  );
+  const gradleVersionOutput = await subProcess.execute('gradle', ['-v'], {});
+  const isGradle3Plus =
```
orsagie, line 447: Seeing as you are calculating it in other places, consider extracting it to its own method?

test/system/multi-module.test.ts
```diff
@@ -415,26 +443,28 @@ test('multi-project-dependency-cycle: scanning the main project works fine', asy
...
+    parseInt(gradleVersionOutput.match(/Gradle (\d+)\.\d+(\.\d+)?/)![1], 10) >=
+    3;
 
-  t.notOk(
-    result.package.dependencies!['com.github.jitpack:subproj'].dependencies![
-      'com.github.jitpack:root-proj'
-    ].dependencies,
-    'dependency cycle is terminated',
-  );
+  // TODO @anthogez, once gradle v2 will be supported on new gradle task, drop this v3+ check
```
orsagie, line 451: will the check also be removed from index? if so, disregard my prev comment

test/system/multi-module.test.ts
anthogez: I can do that, did not because the intention is to deleted it in another pr

test/system/multi-module.test.ts
anthogez: this will be removed in another pr

test/system/multi-module.test.ts
anthogez: this is just temporary specific for a certain gradle version, it will be deleted later and does not default the purpose, as I mentioned on the description in a side pr I will rewrite tests for each case

PR #125, opened 2020-05-28
fix: gradle version cmd issue

anthogez:
- [x] Ready for review

#### What does this PR do?
Fix gradle command version issue

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -374,15 +374,14 @@ function leftPad(s: string, n: number) {
   return ' '.repeat(Math.max(n - s.length, 0)) + s;
 }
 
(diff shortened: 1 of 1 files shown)


orsagie:
Can we have a test that this won't happen again?

anthogez:
> Can we have a test that this won't happen again?

I deleted the check that I add and re-used one that already existed before that have 
these tests https://github.com/snyk/snyk-gradle-plugin/blob/master/test/functional/gradle-version-build-info.test.ts

anthogez:
> Can we have a test that this won't happen again?

https://github.com/snyk/snyk-gradle-plugin/blob/master/test/functional/gradle-version-build-info.test.ts

PR #139, opened 2020-08-12
fix: execute commands relative to wrapper

lili2311:
- [ ] Ready for review

#### What does this PR do?

Execute the commands relative to the wrapper to fix a bug with `--all-projects` where if the command is executed in a dir above a wrapper then gradle fails to execute even if we successfully locate the wrapper.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -484,12 +484,17 @@ async function getAllDeps(
   targetFile: string,
   options: Options,
 ): Promise<JsonDepsScriptResult> {
-  const command = getCommand(root, targetFile);
+  const { command, pathToWrapper } = getCommand(root, targetFile);
+  const relativeRoot = pathToWrapper ? pathToWrapper : root;
+  const targetFileRelative = pathToWrapper
+    ? path.relative(relativeRoot, targetFile)
+    : targetFile;
 
   let gradleVersionOutput = '[COULD NOT RUN gradle -v] ';
+
   try {
     gradleVersionOutput = await subProcess.execute(command, ['-v'], {
-      cwd: root,
+      cwd: relativeRoot,
     });
   } catch (_) {
     // intentionally empty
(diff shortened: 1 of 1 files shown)


lib/index.ts
```diff
@@ -646,25 +659,38 @@ function toCamelCase(input: string) {
...
   pathToWrapper = path.resolve(root, wrapperScript);
   if (fs.existsSync(pathToWrapper)) {
-    return quotLocal + pathToWrapper + quotLocal;
+    return {
+      command: quotLocal + pathToWrapper + quotLocal,
+      pathToWrapper: path.parse(pathToWrapper).dir,
+    };
   }
-  return 'gradle';
+  return { command: 'gradle', pathToWrapper: root };
```
aviadhahami, line 693: why not `pathToWrapper`?

lib/index.ts
```diff
@@ -524,9 +531,15 @@ async function getAllDeps(
...
       const { snykGraph, projectVersion } = extractedJson.projects[projectId];
 
+      // TODO: this should be relative to wrapper
+      // can't be changed without causing naming mismatch to existing projects
+      // may need a migration
       let projectName = path.basename(root);
 
       if (projectId !== extractedJson.defaultProject) {
+        // TODO: this should be relative to wrapper
+        // can't be changed without causing naming mismatch to existing projects
```
aviadhahami, line 541: how can we handle this?

lili2311: let me test locally to confirm this is definitely going to be a problem or not.

lili2311: Doesn't seem to affect the projectID 🎉 Please can you also verify?

lib/index.ts
lili2311: to keep it equivalent to before. Returning this triggers some other logic when a wrapper is found

PR #144, opened 2020-08-28
fix: break edge case cycles

anthogez:
- [x] Ready for review

#### What does this PR do?
break edge case cycles

### DIFF
--- a/lib/find-cycles.ts
+++ b/lib/find-cycles.ts
@@ -1,0 +1,56 @@
+export function findCycles(
+  ancestorsChain,
+  childrenChain,
+  parentId: string,
+  currentNode: string,
+  alreadyVisited: Set<unknown>,
+) {
+  if (ancestorsChain && childrenChain && parentId && currentNode) {
+    const currentAncestors = ancestorsChain.get(parentId);
+    const currentChildren = childrenChain.get(currentNode);
+    if (parentId === currentNode) {
+      return true;
+    }
+
+    if (currentAncestors) {
+      if (
+        currentAncestors.includes(currentNode) ||
+        currentAncestors.includes(parentId)
+      ) {
+        return true;
+      }
+      for (const ancestor of currentAncestors) {
+        if (!alreadyVisited.has(ancestor) && ancestor !== currentNode) {
+          alreadyVisited.add(ancestor);
+          return findCycles(
+            ancestorsChain,
+            childrenChain,
+            ancestor,
+            currentNode,
+            alreadyVisited,
+          );
+        }
+      }
+    } else if (currentChildren) {
+      if (
+        currentChildren.includes(parentId) ||
+        currentChildren.includes(currentNode)
+      ) {
+        return true;
+      }
+      for (const children of currentChildren) {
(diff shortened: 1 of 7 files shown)


lib/find-cycles.ts
```diff
@@ -0,0 +1,56 @@
+export function findCycles(
```
anthogez, line 1: this is temporary, once cycles are allowed it will be deleted

test/system/plugin.test.ts
```diff
@@ -55,7 +55,15 @@ test('multi-config: both compile and runtime deps picked up by default', async (
   // double parsing to have access to internal depGraph data, no methods available to properly
   // return the deps nodeIds list that belongs to a node
   const graphObject: any = JSON.parse(JSON.stringify(result.dependencyGraph));
-  t.ok(graphObject.graph.nodes[0].deps.length === 6, 'top level deps');
+  const GRADLE_VERSION = process.env.GRADLE_VERSION;
+  if (GRADLE_VERSION) {
+    const major = parseInt(GRADLE_VERSION.split('.')[0]);
+    if (major < 4) {
+      t.ok(graphObject.graph.nodes[0].deps.length === 11, 'top level deps');
```
anthogez, line 62: exists a divergence between gradle v4+ and previously versions e.g. v2, v3 in how handle dependencies

test/manual/build-graph.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    };
+
+    // a-> b-> c-> a
+    const projectName = 'goof-project';
+    const projectVersion = 'unspecified';
+    const depGraph = await buildGraph(snykGraph, projectName, projectVersion);
+    const actual = depGraph
+      .getPkgs()
+      .reduce((acc, pkg) => acc + depGraph.countPathsToRoot(pkg), 0);
+    expect(actual).toBeTruthy();
```
gitphill, line 49: lets assert the dep-graph equals expected

test/system/plugin.test.ts
gitphill: this feels wrong, we should detect the same packages and the same version?

lib/find-cycles.ts
```diff
@@ -0,0 +1,56 @@
...
+  ancestorsChain: Map<any, any>,
+  childrenChain: Map<any, any>,
```
gitphill, line 3: any? can we do better?

test/system/plugin.test.ts
anthogez: Added another assert where we verify the pkgs. With this code we are just handling the support of `DAG` here and the divergences that exists between gradle v2,v3 vs v4+

It will be dropped once we allow the cycles

orsagie (approved):
In future, can you split these kinds of PRs to two commits (refactor and fix) so I can see the actual fix more easily 😅

anthogez:
> In future, can you split these kinds of PRs to two commits (refactor and fix) so I can see the actual fix more easily 😅

Sure, thank you! :D

PR #149, opened 2020-10-04
feat: upgrade call graph builder to 1.15.1

juanamari94:
- [X] Ready for review

#### What does this PR do?

Upgrades call graph builder dependency to v1.15.1

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
   "dependencies": {
     "@snyk/cli-interface": "2.9.1",
     "@snyk/dep-graph": "^1.19.4",
-    "@snyk/java-call-graph-builder": "^1.14.0",
+    "@snyk/java-call-graph-builder": "1.15.1",
     "@types/debug": "^4.1.4",
     "chalk": "^3.0.0",
     "debug": "^4.1.1",


package.json
```diff
@@ -47,7 +47,7 @@
   "dependencies": {
     "@snyk/cli-interface": "2.9.1",
     "@snyk/dep-graph": "^1.19.4",
-    "@snyk/java-call-graph-builder": "^1.14.0",
+    "@snyk/java-call-graph-builder": "^1.15.1",
```
darmalovan, line 50: Juan, can we pin it, so removing the caret?

anthogez: More about caret and why we prevent the usage on internal dependencies

**Caret ranges are ideal when an author may make breaking changes between 0.2.4 and 0.3.0 releases, which is a common practice. However, it presumes that there will not be breaking changes between 0.2.4 and 0.2.5. It allows for changes that are presumed to be additive (but non-breaking), according to commonly observed practices.** https://docs.npmjs.com/misc/semver

juanamari94: done

darmalovan (approved):

anthogez (approved):

PR #155, opened 2020-11-23
fix: gradle deps conflict resolution

anthogez:
- [x] Ready for review

#### What does this PR do?

This pr handle each resolvable configuration separately as these will be the containers under which Gradle will resolve it's dependencies. We are no longer merging all the configs and attributes. We are giving priority to configs that can be resolved

### DIFF
--- a/lib/init.gradle
+++ b/lib/init.gradle
@@ -1,6 +1,7 @@
 import groovy.json.JsonOutput
 import java.util.regex.Pattern
 import java.util.regex.Matcher
+import org.gradle.util.GradleVersion
 
 // Snyk dependency resolution script for Gradle.
 // Tested on Gradle versions from 2.14 to 5.4.1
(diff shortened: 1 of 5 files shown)


test/system/plugin.test.ts
```diff
@@ -222,108 +222,6 @@ test('multi-config: only deps for specified conf are picked up (using legacy CLI
   t.ok(graphObject.graph.nodes[0].deps.length === 1, 'top level deps 1');
 });
 
-test('tests for Gradle 3+', async (t0) => {
-  const gradleVersionOutput = await subProcess.execute('gradle', ['-v'], {});
-  const isGradle3Plus =
```
anthogez, line 227: This test is no longer relevant, we are no longer `merging all the configurations` and trying to deal with multi-config from a single merged config source

lib/init.gradle
```diff
@@ -192,67 +193,74 @@ allprojects { everyProj ->
...
                     println('SNYKECHO processing project: ' + proj.name)
-                    def snykConf = null
-                    def mergeableConfs = proj.configurations
-                        .findAll({ it.name != 'snykMergedDepsConf' && it.name =~ confNameFilter })
 
-                    // Drop all the configrations that don't match the attribute filter
-                    if (confAttrSpec != null) {
-                        mergeableConfs = mergeableConfs.findAll(matchesAttributeFilter)
-                    }
+                    def snykConf = null
```
anthogez, line 221: no longer merging all the configurations in a new one `snykMergedDepsConf`

test/system/failure-states.test.ts
```diff
@@ -11,7 +11,7 @@ test('malformed build.gradle', async (t) => {
     inspect(
       '.',
       path.join(fixtureDir('malformed-build-gradle'), 'build.gradle'),
-      { args: ['--configuration', 'compileOnly'] },
+      { args: ['--configuration', 'compileClasspath'] },
```
anthogez, line 14: `compileOnly` just declares the dependency of the application while `compileClasspath` is a `configuration meant to be resolved` - more here https://docs.gradle.org/current/userguide/declaring_dependencies.html

test/system/kotlin.test.ts
```diff
@@ -51,7 +51,21 @@ if (kotlinSupported) {
       const graphObject: any = JSON.parse(
         JSON.stringify(result.dependencyGraph),
       );
-      t.ok(graphObject.graph.nodes[0].deps.length === 8, 'top level deps');
+
+      t.ok(
+        graphObject.graph.nodes[0].deps.length === 3,
```
anthogez, line 56: number has been reduced because we are no longer considering deps that are not completely resolvable

test/system/plugin.test.ts
```diff
@@ -359,7 +257,7 @@ test('custom dependency resolution via configurations* is NOT suppored (known pr
   });
 
   t.ok(
-    nodeIds.indexOf('com.android.tools:annotations@25.3.0') !== -1, // 25.2.0 in NOT forced
+    nodeIds.indexOf('com.android.tools:annotations@25.2.0') !== -1, // 25.2.0 instead of 25.3.0 due of dependency conflict resolution
```
anthogez, line 260: as the comment says we were not resolving the dependency version properly, in practice now if we have a conflict we are going to follow what gradle suggestion

lib/init.gradle
```diff
@@ -192,67 +193,74 @@ allprojects { everyProj ->
...
-
-                        // We use the only matching configuration, with its attributes.
-                        snykConf = mergeableConfs.first()
+                    if(isGradleVersion3Plus) {
+                        // We are looking for a configuration that `canBeResolved`, because it's a configuration for which 
+                        // we can compute a dependency graph and that contains all the necessary information for resolution to happen.
+                        if (confAttrSpec != null) {
+                            // Drop all the configrations that don't match the attribute filter
+                            snykConf = proj.configurations
+                            .findAll({ it.canBeResolved == true && it.canBeConsumed == false && it.name =~ confNameFilter && matchesAttributeFilter(it)  })
```
anthogez, line 234: By default we care only by canBeResolved dependencies and canBeConsumed false, 
if we find 0 results then we are going to look for canBeResolved and canBeConsumed.

But normally is preferable return only canBeResolved and cannotBeConsumed configs

test/system/plugin.test.ts
```diff
@@ -149,8 +149,8 @@ test('multi-config: only deps for specified conf are picked up (fuzzy match)', a
   const result = await inspect(
     '.',
     path.join(fixtureDir('multi-config'), 'build.gradle'),
-    { 'configuration-matching': 'pileon' },
-  ); // case-insensitive regexp matching "compileOnly"
+    { 'configuration-matching': 'pileclass' },
+  ); // case-insensitive regexp matching "compileClasspath"
```
anthogez, line 153: validating regex again

gitphill (approved):

orsagie (approved):

lib/init.gradle
shanman190: @anthogez, this actually could result in seeing less dependencies than intended. It's recommended to only observe `canBeResolved` when interested in dependency containers. While the internal Gradle configurations are correct insofaras `canBeResolved=true, canBeConsumed=false`, plugin authors which add additional configurations and potentially additional dependencies are not so thorough.

By also doing the small simplification you'd be able to remove the second pass case and align with how Gradle does the resolution internally.

PR #161, opened 2021-01-30
fix: scanning lock due of unresolved deps

anthogez:
- Renamed snykConf to projectConfigs, to improve the readability

- Created configsSuccessfullyResolved method that confirms if
  configs sets as canBeResolved=true, can be really resolved or not.

  If there is an error now, while resolving configs set as
  canBeResolved=true, we are no longer blocking the scanning. We ignore
  those configs whose depGraph cannot be computed and move forward with the
  scanning process since it's not a snyk issue but a given gradle project
  resolution config caused by bad config or 3rd party gradle dependencies bad behaving.
  
By going into `test/fixtures/successful-scan-with-unresolved-custom-configs/build.gradle` and running `gradle -q dependencies` you will see the following message (pic below)

This fixture emulates issue gradle/gradle#6854, **where gradle cannot resolve incremental analysis configurations**.

Screen Shot 2021-01-31 at 18 54 04

What does **FAILED** means? Means any dependency belonging to these configuratios failed to be resolved (compute depGraph)

Continue reading about failed resolution in Gradle Docs.. https://docs.gradle.org/current/userguide/viewing_debugging_dependencies.html#example_rendering_the_dependency_report_for_a_custom_configuration

### DIFF
--- a/test/fixtures/successful-scan-with-unresolved-custom-configs/build.gradle
+++ b/test/fixtures/successful-scan-with-unresolved-custom-configs/build.gradle
@@ -1,0 +1,16 @@
+plugins {
+    id 'io.gatling.gradle' version "3.5.0"
+}
+
+build.dependsOn gatlingClasses
+
+dependencies {
+    gatling group: 'org.scalaj', name: 'scalaj-http_2.13', version: '2.4.2'
+}
+
+gatling {
+    simulations = {
+        exclude "**/paxos/BaseSimulation.scala"
+        exclude "**/paxos/Configuration.scala"
+    }
+}
--- a/test/system/multi-module.test.ts
+++ b/test/system/multi-module.test.ts
@@ -299,16 +299,6 @@ test('single-project: array of one is returned with allSubProjects flag', async 
   t.ok(nodeIds.indexOf('commons-httpclient:commons-httpclient@3.1') !== -1);
 });
 
-test('multi-project-some-unscannable: allSubProjects fails', async (t) => {
-  await t.rejects(
-    inspect(
-      '.',
-      path.join(fixtureDir('multi-project-some-unscannable'), 'build.gradle'),
-      { allSubProjects: true },
-    ),
-  );
-});
-
 test('multi-project-some-unscannable: gradle-sub-project for a good subproject works', async (t) => {
   const options = {
     subProject: 'subproj ',
--- a/lib/init.gradle
+++ b/lib/init.gradle
@@ -4,7 +4,7 @@ import java.util.regex.Matcher
 import org.gradle.util.GradleVersion
 
 // Snyk dependency resolution script for Gradle.
-// Tested on Gradle versions from 2.14 to 5.4.1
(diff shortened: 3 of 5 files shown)


test/system/multi-module.test.ts
```diff
@@ -299,14 +299,18 @@ test('single-project: array of one is returned with allSubProjects flag', async
...
+  const result = await inspect(
+    '.',
+    path.join(fixtureDir('multi-project-some-unscannable'), 'build.gradle'),
+    { allSubProjects: true },
   );
+
+  t.deepEqual(result.plugin.meta!.allSubProjectNames, [
+    'root-proj',
+    'subproj',
+    'subproj-fail',
```
anthogez, line 312: Scanning process will no longer be stuck/fail due of unresolved configs, `subproj-fail` will be empty, and that's fine until I raise another pr where I will decide its destiny 

n.b. it would fail anyway when using 'gradle -q dependencies'

test/system/multi-module.test.ts
```diff
@@ -299,15 +299,15 @@ test('single-project: array of one is returned with allSubProjects flag', async
...
-});
+// test('multi-project-some-unscannable: allSubProjects fails', async (t) => {
+//   await t.rejects(
+//     inspect(
+//       '.',
+//       path.join(fixtureDir('multi-project-some-unscannable'), 'build.gradle'),
+//       { allSubProjects: true },
+//     ),
+//   );
+// });
```
anthogez, line 310: There will be a new test that makes sense to replace this one

anthogez, line 310: This test will be replaced with a test that makes more sense with the new logic on deps resolution check.

test/system/multi-module.test.ts
```diff
@@ -299,16 +299,6 @@ test('single-project: array of one is returned with allSubProjects flag', async
   t.ok(nodeIds.indexOf('commons-httpclient:commons-httpclient@3.1') !== -1);
 });
 
-test('multi-project-some-unscannable: allSubProjects fails', async (t) => {
```
anthogez, line 302: Was replaced with a test that makes more sense with the new logic on deps resolution check.
Here we were rejecting, because in our previously behaviour we wanted the whole scanning to fail if a single sub-module suddenly failed due of a configuration that could not be resolved (could not compute dep-graph), which is a scenario that can happen when users are using 3rd party dependencies that are not well handled and do not support resolution, either have bad a config to a private registry

test/fixtures/successful-scan-with-unresolved-custom-configs/build.gradle
```diff
@@ -0,0 +1,16 @@
+plugins {
+    id 'io.gatling.gradle' version "3.5.0"
```
anthogez, line 2: This fixture emulates the issue below where gradle cannot resolve `incremental analysis configurations`
https://github.com/gradle/gradle/issues/6854

gitphill (approved):

PR #162, opened 2021-02-01
fix: more than one line with jsondeps

anthogez:
- swallowing unresolvable configs error msg in order to prevent mass spam in
script output while scanning projects with large graph output to prevent parse errors...

### DIFF
--- a/lib/init.gradle
+++ b/lib/init.gradle
@@ -136,7 +136,7 @@ def configsSuccessfullyResolved(configurations) {
             configuration.resolve();
             resolvedConfigurations.add(configuration);
            } catch(Exception ex) {    
-            println('NOT_RESOLVED ' + ex.toString())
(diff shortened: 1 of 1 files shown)


lib/init.gradle
```diff
@@ -136,7 +136,7 @@ def configsSuccessfullyResolved(configurations) {
             configuration.resolve();
             resolvedConfigurations.add(configuration);
            } catch(Exception ex) {    
-            println('NOT_RESOLVED ' + ex.toString())
+             // Swallowing it
```
ekbsnyk, line 139: Is it safe to swallow this and not log anything? E.g. do we handle unresolved configs properly elsewhere, e.g. inside configuration.resolve()?

ekbsnyk (approved):

PR #168, opened 2021-02-23
fix: make custom gradle init script parameter explicit

juanamari94:
- [x] Ready for review

#### What does this PR do?

The `--init-script=path` argument was being set as a raw argument in the CLI, but recently due to the addition of the `--initScript` in gradle for the reachability feature this has been made more explicit. In order to stay true to this convention and make it more evident to users through documentation, we're making the `initScript` option explicit and being sent as a parameter for the `getAllDeps` function.

`snyk test --file=build.gradle --init-script=init.gradle`

The initialization script is part of gradle and was not being sent explicitly at all for dependency discovery. This fixes it.

https://snyksec.atlassian.net/browse/FLOW-666


pwnslinger (approved):
Thanks for applying the `--init-script` modifications to other portions of the code. When I was testing on our codebase, I noticed a similar error in connection to our internal artifactory. As long as you cross-referenced for all the places we have `getCommand()` and passed the `initScript` we should be fine. Here is the relevant issue on this:

https://support.snyk.io/hc/en-us/requests/8839?page=1 

I am looking for the merge since it looks fine to me. I was about to apply similar patches in my upcoming PR.

juanamari94:
Hey @pwnslinger , nice to see you again!

The code is mostly done. I'm trying to handle some cases since our tests run with different versions of npm and gradle and an edge case came up. Hopefully we'll be getting this merged soon!

darmalovan (approved):

test/fixtures/with-init-script/init.gradle
```diff
@@ -0,0 +1,8 @@
+import java.text.SimpleDateFormat
+gradle.allprojects{
+  ext.buildTime = {
+   def df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'")
+   df.setTimeZone(TimeZone.getTimeZone("UTC"))
+   return df.format(new Date())
```
anthogez, line 6: What's the purpose of this fixture? I understand, it's being triggered but how helpful it would be for CLI Scanning/Reachable?

test/fixtures/no wrapper/init.gradle
```diff
@@ -0,0 +1 @@
+println "I'm an init file"
```
anthogez, line 1: What's the purpose of this fixture? I understand, it's being triggered but how helpful it would be for CLI Scanning/Reachable?

anthogez, line 1: Same question concern of https://github.com/snyk/snyk-gradle-plugin/pull/168/commits/f1dc776df2c1f2f69e512bca4f40b111aed9d95e#r582976810

test/fixtures/with-init-script/init.gradle
juanamari94: The point of this fixture is entirely unrelated to scanning/reachable. The point of this `init.gradle` is to create a dependency between this file and the `build.gradle`, such that gradle won't work without both of them.

This is a regression test, it's making sure that given both an init file (which is required by the `build.gradle`) and the `build.gradle` itself, the gradle commands will not fail.

test/fixtures/no wrapper/init.gradle
juanamari94: This is simply an init file with no dependencies whatsoever for certain tests that use initialization files (e.g: reachability) to make sure that the gradle commands work correctly with some arbitrary init file.

anthogez (approved):

pwnslinger:
This patch is faulty and is not working. You didn't need to introduce a new `initScriptPath` and pass it all the way over to `getAllDepsOneProject` or `getAllDepsAllProjects`. I already added `init-script` to the `Options` in my previous PR! Look at this please:

```ts
// snyk-gradle-plugin/lib/index.ts#L70
type Options = api.InspectOptions & GradleInspectOptions;
```

and introduced here: 
https://github.com/snyk/snyk/commit/859f71d5a11ac1eb1b5d51e8766421ee92bc4ef1#diff-cb98b6b0e8336713b5d56cb8986db5a61310c740a8317b9bf42bc90be49e0fb8

and above in the same file under `GradleInspectOptions`:

```ts
export interface GradleInspectOptions {

  'configuration-matching'?: string;

  'configuration-attributes'?: string;

  daemon?: boolean;
  reachableVulns?: boolean;
  initScript?: string;
}
```

when you do like that, you should change all of your codebase to taint the argument all the way from snyk/cli to here in your snyk-gradle-plugin.

PR #172, opened 2021-03-29
feat: use break cycles logic of depgraph lib

anthogez:
- Deleted findCycles and breakCycles logic from snyk-gradle-plugin
  - Multi-config fixture plugin.test.ts has top level deps amount number
    aligned and with correct count, we were appending transitive deps as
top to avoid loosing them due of lack of cycle/break support in
dep-graph lib.
  
  - Dependency **org.jetbrains.kotlin:kotlin-stdlib** is no longer available as
  direct dep in **kotlin.test.ts** (it was wrong) as it's a transitive  dependency see screenshot below.
  It was being `appended` as top due of the lack of cycle/break support in dep-graph lib
  
  dependencies

  - Added @snyk/dep-graph changes

    v1.27.1 remove cache in pathsToRoot

    v1.27 break cyclic graphs when converting into tree

    v1.26 support cycles in pkgPathsToRoot and countPathsToRoot


test/system/kotlin.test.ts
```diff
@@ -61,7 +61,6 @@ if (kotlinSupported) {
         graphObject.graph.nodes[0].deps,
         [
           { nodeId: 'org.jetbrains.kotlin:kotlin-stdlib-jdk8@1.3.21' },
-          { nodeId: 'org.jetbrains.kotlin:kotlin-stdlib@1.3.21' },
```
anthogez, line 64: see GH notes

test/system/plugin.test.ts
```diff
@@ -162,16 +161,8 @@ test('multi-config: both compile and runtime deps picked up by default', async (
   // double parsing to have access to internal depGraph data, no methods available to properly
   // return the deps nodeIds list that belongs to a node
   const graphObject: any = JSON.parse(JSON.stringify(result.dependencyGraph));
-  const gradleVersionOutput = await subProcess.execute('gradle', ['-v'], {});
-  const isGradleVersionLessThan4 =
-    parseInt(gradleVersionOutput.match(/Gradle (\d+)\.\d+(\.\d+)?/)![1], 10) <
-    4;
-
-  if (isGradleVersionLessThan4) {
-    t.ok(graphObject.graph.nodes[0].deps.length === 16, 'top level deps');
```
anthogez, line 171: see GH notes

gitphill (approved):
cc @admons 🎉

anthogez:
> cc @admons 🎉

I forgot of delete this https://github.com/snyk/snyk-gradle-plugin/blob/master/lib/init.gradle#L91 - sending a new push

test/system/kotlin.test.ts
```diff
@@ -4,17 +4,12 @@ import { test } from 'tap';
...
-const GRADLE_VERSION = process.env.GRADLE_VERSION;
-if (GRADLE_VERSION) {
-  const major = parseInt(GRADLE_VERSION.split('.')[0]);
-  if (major < 5) {
-    kotlinSupported = false;
-  }
-}
+const gradleVersionFromProcess = process.env.GRADLE_VERSION;
+const gradleVersionInUse = parseInt(gradleVersionFromProcess?.split('.')[0]);
+const isKotlinSupported = gradleVersionInUse < 5 ? false : true;
```
anthogez, line 9: code clean up

PR #175, opened 2021-04-16
fix: prevent EBADF on macOS while installed with Yarn

JackuB:
Usage of filestream drops file descriptor, thus breaking Snyk CLI installed through Yarn running on macOS.


ekbsnyk (approved):

JackuB:
Updated the commit message and will merge this

PR #176, opened 2021-04-19
fix: specify gradle version in windows tests

Avishagp:
- [X] Ready for review

#### What does this PR do?
Following failing Windows tests in this PR: 
https://github.com/snyk/snyk-gradle-plugin/pull/175

Windows tests are using the latest Gradle version, because the chocolatey installation doesn't specify another version. 
The tests are using Gradle 7 which is not supported causing the tests to fail for Windows.
This change specifies a version.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,7 +54,7 @@ commands:
   install_gradle_windows:
     description: Install gradle
     steps:
-      - run: choco install gradle
+      - run: choco install gradle --version=6.0
   install_sdkman:
     description: Install SDKMAN
     steps:


.circleci/config.yml
```diff
@@ -54,7 +54,7 @@ commands:
   install_gradle_windows:
     description: Install gradle
     steps:
-      - run: choco install gradle
+      - run: choco install gradle --version=6.0
```
anthogez, line 57: We will add more steps where we execute tests for gradle v2 & v5  on windows - but it's going to be a Tardis job - thanks @Avishagp! 🎉

anthogez (approved):

PR #181, opened 2021-04-29
fix: scanning failure caused by missing proj-name

anthogez:
**fix:**

- While processing snykGraphs from extractedJSON to depGraph,
  projectName is generated using the basename of the rootDir,  
  we were not handling it if by any chance it was null or undefined.
  Now that we handling it properly, we are preventing possible scanning errors.
  
**refactor:**

- Clarifying the purpose of few functions

- Moving the loop where we process all
  the existing projects in the extractedJSON from snykGraph to depGraph
  away from getAllDeps to its own function `processProjectsInExtractedJSON`,
  makes the projects process testable, enabling us of have more confidence on expected results

- Added a test for `processProjectsInExtractedJSON`

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -304,12 +304,11 @@ async function getAllDepsOneProject(
   const allSubProjectNames = allProjectDeps.allSubProjectNames;
 
   return subProject
-    ? getSubProject(root, subProject, allProjectDeps, allSubProjectNames)
-    : getRootProject(root, allProjectDeps, allSubProjectNames);
+    ? getSubProject(subProject, allProjectDeps, allSubProjectNames)
+    : getRootProject(allProjectDeps, allSubProjectNames);
 }
 
 function getSubProject(
-  root: string,
   subProject: string,
   allProjectDeps,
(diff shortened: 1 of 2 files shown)


lib/index.ts
```diff
@@ -304,12 +304,11 @@ async function getAllDepsOneProject(
   const allSubProjectNames = allProjectDeps.allSubProjectNames;
 
   return subProject
-    ? getSubProject(root, subProject, allProjectDeps, allSubProjectNames)
```
anthogez: deleted unused params

lib/index.ts
```diff
@@ -304,12 +304,11 @@ async function getAllDepsOneProject(
...
-    : getRootProject(root, allProjectDeps, allSubProjectNames);
+    ? getSubProject(subProject, allProjectDeps, allSubProjectNames)
+    : getRootProject(allProjectDeps, allSubProjectNames);
```
anthogez, line 308: deleted unused `root` param from `getSubProject` and `getRootProject`

mika-bar (approved):

PR #191, opened 2021-10-27
feat: improve gradle plugin accuracy

muscar:
- [x] Ready for review

#### What does this PR do?

This patch set improves the accuracy of the gradle plugin.

Newer version of Gradle added two attributes to configs for controlling
the resolution process: `canBeResolved` and `canBeConsumed` (see
https://docs.gradle.org/current/userguide/declaring_dependencies.html#sec:resolvable-consumable-configs
for details). A recent change to the plugin brought it in line with
Gradle best practices, i.e. prefer `canBeResolved=true` and
`canBeConsumed=false`. If no configs match the criteria, it falls back
to configs with both attributes set to `true`. While this is a good
approach, the implementation is a bit too strict: if _any_ `true`-
`false` configs are found, the `true`-`true` ones won't be considered.
This misses some configs, and hence some dependencies that users expect
to see.

The first change is to relax the strategy, and accept all the configs
that have `canBeResolved=true`. This is equivalent to taking the union
of `true`-`false` and `true`-`true` configs. 

The second change fixes a bug in the plugin where sub-projects with the
same name at different levels of a multi-project hierarchy would
overwrite each other. To patch now uses the full path of the sub-project
to disambiguate between sub-projects with the same name.

In addition to the above changes, the patch refactors the plugin code.

The `init.gradle` file.

Pull the plugin locally, change your local copy of `snyk` to use this version of the plugin, run it on a gradle project.

#### Screenshots

Before:

Screenshot 2021-10-27 at 10 54 45
Screenshot 2021-10-27 at 10 55 01

After:

Screenshot 2021-10-27 at 10 56 38
Screenshot 2021-10-27 at 10 56 49


anthogez (commented):
Nice work @muscar! I will review it soon, just to check. Have we validated that we are not having variants conflicts during dependency resolution by enabling the coexistence of `canBeConsumed=False + canBeResolved=True` & `canBeConsumed=True + canBeResolved=True`? Because that was the primary reason that I've kept them separated when I've introduced the support checking on "canBeResolved & canBeConsumable" thanks! See https://docs.gradle.org/current/userguide/variant_model.html

anthogez (commented):
Would recommend to take a look at this https://github.com/snyk/snyk-gradle-plugin/pull/161

See issue: https://github.com/gradle/gradle/issues/6854

image

anthogez (commented):
Regarding
```
The second change fixes a bug in the plugin where sub-projects with the
same name at different levels of a multi-project hierarchy would
overwrite each other. To patch now uses the full path of the sub-project
to disambiguate between sub-projects with the same name.
```
This should exists in a separated pull request just in case we want to revert the changes on canBeConsumed/CanBeResolved

anthogez (approved):
Approving albeit please create tickets and address what was suggested here in the next days.. thanks!

dtrunley-snyk (approved):

elson-snyk (approved):

PR #198, opened 2022-04-29
chore: update node 8 to 16

magdziarek:
- [X] Ready for review

#### What does this PR do?
In preparation for the Gradle plugin improvements this:
- updates node version from 8 to 16 😨
- updates documentation including changing formatting of the README file (result of running `npm run lint`)

1. Clone and test the following Gradle projects with snyk CLI: [vuln project 1](https://github.com/snyk-fixtures/gradle-vuln-upgrade), [vuln project 2](https://github.com/snyk-fixtures/simple-gradle-repo), [non-vuln Android project](https://github.com/snyk-fixtures/gradle-android-projects), [non-vuln project]( https://github.com/snyk-fixtures/gradle-multi-sub-projets)
2. Check the CLI returns the same results for the above projects when using this branch of snyk-gradle-plugin. To do this:
- Pull this branch of snyk-gradle-plugin 
- Clone snyk/cli
- On the master branch, in the CLI package.json change `"snyk-gradle-plugin": "3.17.0"` to  `"snyk-gradle-plugin": "../snyk-gradle-plugin"`
- Run: `npm i && npm run build`
- Use CLI locally by adding an alias snyk-local='SNYK_API="https://app.dev.snyk.io/api/v1" node /Users/XXXX/cli/dist/cli' -> pointing to your local CLI repo
- Run `snyk-local test` with: [vuln project 1](https://github.com/snyk-fixtures/gradle-vuln-upgrade), [vuln project 2](https://github.com/snyk-fixtures/simple-gradle-repo), [non-vuln Android project](https://github.com/snyk-fixtures/gradle-android-projects), [non-vuln project]( https://github.com/snyk-fixtures/gradle-multi-sub-projets)

[TARDIS-830](https://snyksec.atlassian.net/browse/TARDIS-830)


.npmrc
```diff
@@ -1 +1 @@
-package-lock=false
+package-lock=true
```
gitphill, line 1: we don't need a lockfile in Snyk CLI plugins, let's keeps this set to false

package.json
```diff
@@ -54,5 +54,9 @@
     "debug": "^4.1.1",
     "tmp": "0.2.1",
     "tslib": "^2.0.0"
+  },
+  "engines": {
+    "node": "<=16",
```
gitphill, line 59: less than? I think we want 

```json
"engines": {
    "node": "^16",
    "npm": "^8"
}
```

which means you must use Node.js 16 and above

README.md
```diff
@@ -23,36 +23,37 @@ Please refer to the [Snyk for Java](https://support.snyk.io/hc/en-us/articles/36
...
-| 10    | ✅          |
-| 12    | ✅          |
+| Node | Supported |
+| ---- | --------- |
+| 6    | ✅        |
+| 8    | ✅        |
+| 10   | ✅        |
+| 12   | ✅        |
+| 14   | ❓        |
+| 16   | ✅        |
```
gitphill, line 41: we do have a test for 14 https://github.com/snyk/snyk-gradle-plugin/blob/master/.circleci/config.yml#L182 lets add one for 16 now too?

package.json
magdziarek: I thought it means that we do support 16 and everything below like it says in the README? I read this doc before adding it: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines

magdziarek: > this field is advisory only and will only produce warnings when your package is installed as a dependency.

gitphill: I think the following PRs you're working on demand Node.js 16? In which case we want to warn developers when Node.js 16 is not being used to build this plugin. For example if you use Node.js 14 `npm install` will fail with `ENOTSUP` - `Unsupported engine` which is good if there's something in this repo that requires Node.js 16 to be built correctly.

magdziarek: What about the README? It says we support <=16

gitphill: yea I think that's talking about runtime (what you run the Snyk CLI with) as opposed to build time here. I'm looking at the more recent CLI dependencies https://github.com/snyk/cli/blob/master/packages/snyk-protect/package.json#L23 and they have `>=10` here, maybe that what we should be using. Lets check with team hammer.

README.md
magdziarek: This is a test that will only run in CircleCi, is that right?

gitphill: yep, we have Unix and Windows test with various combinations of versions for Gradle, Java and Node.js

package.json
magdziarek: This says Snyk CLI requires min version 12: https://support.snyk.io/hc/en-us/articles/360004183317-What-version-of-Node-is-required-for-Snyk- and ending Node 10 support: https://updates.snyk.io/ending-snyk-cli-support-for-node-js-v10-219175. This is runtime, like you said, right? What should I ask Hammer? If Snyk CLI plugins should still support node version 8 at build time?

PR #202, opened 2022-05-05
chore: upgrade jest and related deps

magdziarek:
- [X] Ready for review

#### What does this PR do?
- Upgrades versions of: jest, ts-jest and @types/jest

- In android.spec.ts everything in the describe block should be skipped on default. With the older versions it was. Now, the describe.skip() means tests will be skipped but [the describe block will still be run](https://jestjs.io/docs/api#describeskipname-fn) and it will fail if the environment variable is not set. Therefore, this adds a separate test block to check if the env var exists.

Run `npm run test-manual && npm run test-functional`
(`npm run test` will also run tap test which aren't relevant here)

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -12,7 +12,7 @@ module.exports = {
   ], // Remove when all tests are using Jest
   globals: {
     'ts-jest': {
-      tsConfig: '<rootDir>/test/manual/tsconfig.json',
+      tsconfig: '<rootDir>/test/manual/tsconfig.json',
     },
   },
   modulePathIgnorePatterns: ['<rootDir>/test/.*fixtures'],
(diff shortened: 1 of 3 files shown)


jest.config.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
   ], // Remove when all tests are using Jest
   globals: {
     'ts-jest': {
-      tsConfig: '<rootDir>/test/manual/tsconfig.json',
+      tsconfig: '<rootDir>/test/manual/tsconfig.json',
```
gitphill, line 15: not for this PR but you can write jest config as Typescript module these days https://jestjs.io/docs/configuration may help getting the types correct

gitphill (approved):

jest.config.js
magdziarek: Created a ticket for this https://snyksec.atlassian.net/browse/TARDIS-911

magdziarek:
added explanation on how to run the skipped manual test in android.spec.ts

PR #203, opened 2022-05-05
chore: update README

magdziarek:
- [X] Ready for review
- [X] Follows CONTRIBUTING rules

#### What does this PR do?
Snyk CLI supports node versions  >=12. Also, with newer versions of dependencies, CircleCI tests fail for node 8 and 10. Jest is one example: https://jestjs.io/blog/2021/05/25/jest-27#miscellaneous-breaking-changes

PR #204, opened 2022-05-05
feat: add tests for supported node versions

magdziarek:
- [X] Ready for review

#### What does this PR do?
Adds tests for different Node versions following [this PR ](https://github.com/snyk/snyk-gradle-plugin/pull/200)

Next related [PR](https://github.com/snyk/snyk-gradle-plugin/pull/203)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,6 +22,19 @@ windows_defaults: &windows_defaults
   executor:
     name: win/default
 
+test_matrix_unix: &test_matrix_unix
+  matrix:
+    parameters:
+      node_version: ['12', '16']
+      jdk_version: ['8.0.292.j9-adpt']
+      gradle_version: ['3.4.1', '4.10', '5.5', '6.2.1']
+test_matrix_win: &test_matrix_win
+  matrix:
+    parameters:
+      node_version: ['12', '16']
+      jdk_version: ['8']
+      gradle_version: ['4.10', '5.5', '6.2.1']
+
 commands:
   install_deps:
     description: Install dependencies
@@ -53,8 +66,12 @@ commands:
           command: npm --version
   install_gradle_windows:
     description: Install gradle
+    parameters:
+      gradle_version:
+        type: string
+        default: ''
     steps:
-      - run: choco install gradle --version=6.0
+      - run: choco install gradle --version=<< parameters.gradle_version >>
   install_sdkman:
     description: Install SDKMAN
     steps:
@@ -108,7 +125,6 @@ jobs:
       - run:
           name: Run lint
           command: npm run lint
-
   test-windows:
     <<: *defaults
     <<: *windows_defaults
@@ -121,7 +137,8 @@ jobs:
           node_version: << parameters.node_version >>
       - install_jdk_windows:
           jdk_version: << parameters.jdk_version >>
-      - install_gradle_windows
+      - install_gradle_windows:
+          gradle_version: << parameters.gradle_version >>
       - install_deps
       - show_node_version
       - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -332,6 +344,72 @@ workflows:
...
+          name: Windows Tests for JDK=8 Node=10
+          context: nodejs-install
+          jdk_version: '8'
+          node_version: '10.21.0'
+          requires:
+            - Lint
+          filters:
+            branches:
+              ignore:
+                - master
```
gitphill, line 412: I find CircleCI matrix easier to read https://circleci.com/blog/circleci-matrix-jobs/
there's an example in the snyk-sbt-plugin https://github.com/snyk/snyk-sbt-plugin/blob/master/.circleci/config.yml#L18

magdziarek: 1. Using the matrix turned out to be a bit tricky because of Java and Gradle compatibility.  https://docs.gradle.org/current/userguide/compatibility.html -> shouldn't this be the same in our README? 
2. When I ran tests in CircleCI first time, Windows passed even when Gradle=3 or 4 and JDK= 11 (contrary to the above compatibility matrix). On another run the same tests failed erroring when installing Node on Windows "Too long with no output (exceeded 10m0s): context deadline exceeded"...
3. What is a reasonable number of tests to have here? For example the SBT plugin tests only one SBT version and only on Windows. Here, testing only Gradle 6 with different JDK and Node versions takes over 10 min

magdziarek: @gitphill we discussed it in Beta Tov yesterday but haven't made any decisions - what do you think about the above?

muscar:
Duration: 11m 26s. Not too bad.

Just out of curiosity: does testing gradle 7 fail?

magdziarek:
> Duration: 11m 26s. Not too bad.
> 
> Just out of curiosity: does testing gradle 7 fail?

@muscar it only fails one of the tap system tests https://app.circleci.com/pipelines/github/snyk/snyk-gradle-plugin/406/workflows/e304de72-e9b5-407a-b058-c69a701b9d2e/jobs/4439 - it's not difficult to fix but I'd rather do it in another PR after converting tests to jest are approved

Jdunsby (approved):

.circleci/config.yml
```diff
@@ -22,6 +22,19 @@ windows_defaults: &windows_defaults
...
+    parameters:
+      node_version: ['12', '16']
+      jdk_version: ['8.0.292.j9-adpt']
+      gradle_version: ['3.4.1', '4.10', '5.5', '6.2.1']
+test_matrix_win: &test_matrix_win
+  matrix:
+    parameters:
+      node_version: ['12', '16']
+      jdk_version: ['8']
+      gradle_version: ['4.10', '5.5', '6.2.1']
```
gitphill, line 36: no 3 on windows?

magdziarek: Nope. It fails with 3. It wasn't included in previous tests on windows.

gitphill: ok!

gitphill (approved):

.circleci/config.yml
magdziarek: https://app.circleci.com/pipelines/github/snyk/snyk-gradle-plugin/445/workflows/4d56635b-eddd-4762-809a-35178897c4b4/jobs/5103 this is how it fails

magdziarek: It fails on the Kotlin test - in Gradle 3 Kotlin wasn't supported. However - why does it not fail on Unix is what I am wondering....

PR #208, opened 2022-05-24
feat: make --gradle-accept-legacy-config-roles behaviour the default

Jdunsby:
- [ ] Documentation written
- [x] Commit history is tidy

### What this does

This change turns the functionality behind the `--gradle-accept-legacy-config-roles` flag into default behaviour. It also removes the legacy init script and updates tests accordingly.

A comment from the repo:
To follow best practices for newer versions of Gradle, avoid configs that are both resolvable _and_ consumable, i.e. `canBeResolved == true && canBeConsumed == true`. Legacy projects will have these, so CLI scan might report fewer dependencies than the users expect. This flag allows the users to relax the configuration filtering strategy. This runs the risk of configuration variant conflicts, but the user should know what they get themselves into, so we may need to announce this ahead of release. 
See: https://docs.gradle.org/current/userguide/declaring_dependencies.html#sec:resolvable-consumable-configs

### More information

- [Jira ticket TARDIS-923](https://snyksec.atlassian.net/browse/TARDIS-923)


gitphill:
LGTM - open up 'ready for review'?

gitphill (approved):

PR #209, opened 2022-06-08
feat: prune across the dep-graph

gitphill:
- [X] Documentation written
- [X] Commit history is tidy

### What this does

Before this change we have been traversing all transitive lines as deep as possible creating at times very large graphs (many edges).

This can cause us difficulties performing vulnerable path calculations when considering all paths from all nodes to the root.

By pruning already seen nodes we can reduce the problem space dramatically.

To achieve this first perform a breadth first search over the result of init.gradle.
Checking whether each node as been seen before queuing it's children.
If a node has been seen simply mark it as pruned and do not queue it's children terminating the line early.

It is safe to discard these duplicate transitive lines because there can only be one version of a package on the classpath for Java projects.
There are some situations when Gradle considers higher versions of deeper duplicate packages winners. We're relying on init.gradle to resolve these situations.

When presenting remediation (upgrades) Snyk will attempt to find and fix the effective (non-pruned) lines. This is correct as it's those lines that should matter above all others.


lib/graph.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export function findChildren(
+  parentId: string,
+  snykGraph: SnykGraph,
+): QueueItem[] {
+  const result: QueueItem[] = [];
+  for (const id of Object.keys(snykGraph)) {
+    const node = snykGraph[id];
+    if (node?.parentIds?.indexOf(parentId) > -1) {
```
magdziarek, line 69: What is this check? Can a node have multiple parents?

gitphill: This checks the node in the loop is a child of the given parent.

You can have multiple parentIds. For example if you have two top level deps A and B and they both depend transitively on C, C would indicate A and B are parents. This is still true in the 'snykGraph' object (the result of running init.gradle).

We're changing the way we represent that in the dep-graph though. In the A, B, C example, and new node 'C:pruned' gets created to indicate the C seen under B is the second time we've seen it so we terminate the line and go no deeper, but we allow the C under A to traverse as normal. Effectively making the dep-graph more directional than the 'snykGraph'.

test/system/plugin.test.ts
```diff
@@ -196,3 +198,16 @@ test('custom dependency resolution via configurations* is NOT supported (known p
     nodeIds.indexOf('com.android.tools:annotations@25.2.0'),
   ).toBeGreaterThanOrEqual(-1); // 25.2.0 instead of 25.3.0 due of dependency conflict resolution
 });
+
+test('repeated transitive lines terminated at duplicate node and labeled pruned', async () => {
+  const pruneSpringBeans = fixtureDir('pruned-spring-app');
```
magdziarek, line 203: How about changing pruneSpringBeans to pathToFixture?

gitphill: yep can do

lib/graph.ts
Jdunsby, line 69:
```suggestion
    if (node?.parentIds?.includes(parentId)) {
```

gitphill: no 0 is still of interest isn't it?

gitphill: oh sorry missed the includes - don't think we can use it without changing Typescript

gitphill: I get 

```
lib/graph.ts:69:26 - error TS2339: Property 'includes' does not exist on type 'string[]'.

69     if (node?.parentIds?.includes(parentId)) {
```

inclined to leave this for another day

gitphill: oh it's `tsconfig-test.json` - i'll change this to align with `tsconfig.json` which set `es2017`

gitphill: https://github.com/snyk/snyk-gradle-plugin/pull/209/commits/8fb3dd7b13d28437270e87476b21790e1737a627

Jdunsby: Sorry, didn't mean to cause unnecessary work here.

Jdunsby (approved):

lib/graph.ts
gitphill: good catch - we could do with removing tsconfig-test.json at some point I don't think its necessary

PR #211, opened 2022-06-10
chore: change jest config file to typescript

magdziarek:
- [X] Commit history is tidy

### What this does
We want to get rid of the leftover javascript config file because jest allows typescript config files. 

### More information
- [TARDIS-911](https://snyksec.atlassian.net/browse/TARDIS-911)


jest.config.ts
```diff
@@ -13,3 +15,4 @@ module.exports = {
   },
   modulePathIgnorePatterns: ['<rootDir>/test/.*fixtures'],
 };
+export default config;
```
gitphill, line 18: https://jestjs.io/docs/configuration

the exported object should be a async function (returning a Promise)

gitphill: oh sorry no it can be either

gitphill (approved):

PR #215, opened 2022-07-05
feat: create-new-test-fixtures-for-gradle-7

Jdunsby:
- [x] Documentation written
- [x] Commit history is tidy

### What this does

Adds tests to demonstrate support for Gradle 7 specific features. 

- Jira ticket: [TARDIS-932](https://snyksec.atlassian.net/browse/TARDIS-932)


something.json
```diff
@@ -0,0 +1 @@
...arams@5.8.2"},{"nodeId":"org.junit.jupiter:junit-jupiter-api@5.8.2"},{"nodeId":"org.junit:junit-bom@5.8.2"}]},{"nodeId":"org.junit.jupiter:junit-jupiter-params@5.8.2","pkgId":"org.junit.jupiter:junit-jupiter-params@5.8.2","deps":[{"nodeId":"org.junit.jupiter:junit-jupiter-api@5.8.2"},{"nodeId":"org.junit:junit-bom@5.8.2"},{"nodeId":"org.apiguardian:apiguardian-api@1.1.2"}]},{"nodeId":"org.junit.jupiter:junit-jupiter-api@5.8.2","pkgId":"org.junit.jupiter:junit-jupiter-api@5.8.2","deps":[{"nodeId":"org.junit.platform:junit-platform-commons@1.8.2"},{"nodeId":"org.junit:junit-bom@5.8.2"},{"nodeId":"org.apiguardian:apiguardian-api@1.1.2"},{"nodeId":"org.opentest4j:opentest4j@1.2.0"}]},{"nodeId":"org.junit.platform:junit-platform-commons@1.8.2","pkgId":"org.junit.platform:junit-platform-commons@1.8.2","deps":[{"nodeId":"org.junit:junit-bom@5.8.2"},{"nodeId":"org.apiguardian:apiguardian-api@1.1.2"}]},{"nodeId":"org.junit:junit-bom@5.8.2","pkgId":"org.junit:junit-bom@5.8.2","deps":[]},{"nodeId":"org.apiguardian:apiguardian-api@1.1.2","pkgId":"org.apiguardian:apiguardian-api@1.1.2","deps":[]},{"nodeId":"org.opentest4j:opentest4j@1.2.0","pkgId":"org.opentest4j:opentest4j@1.2.0","deps":[]}]}}
```
gitphill, line 1: What's this

README.md
```diff
@@ -37,16 +37,14 @@ Please refer to the [Snyk for Java](https://docs.snyk.io/products/snyk-open-sour
 | 14   | ✅        |
 | 16   | ✅        |
```
gitphill, line 38: are we saying we support JDK17 now? (it's in the circleci config)

tsconfig-test.json
```diff
@@ -1,20 +1,16 @@
...
-      "module": "commonjs",
-      "allowJs": false,
-      "sourceMap": true,
-      "strict": false,
-      "declaration": true,
-      "importHelpers": true,
-      "typeRoots": ["./node_modules/@types", "./typings"],
+    "outDir": "./dist-test",
+    "pretty": true,
+    "target": "es2017",
```
gitphill, line 5: was this necessary for this PR, or just something we noticed could do with upgrading?

gitphill: oh I think a rebase will remove this diff...

something.json
Jdunsby: A careless mistake which shall be removed swiftly 🗑️

test/system/fixtures-with-wrappers.test.ts
```diff
@@ -0,0 +1,28 @@
...
+    test(`${dsl} fixture: ${fixtureName}`, async () => {
+      const fixturePath = getPathToFixture(fixtureName);
+      const buildFileName = isKotlin ? 'build.gradle.kts' : 'build.gradle';
+      const pathToBuildConfig = path.join(fixturePath, buildFileName);
+      const expectedDepGraph = require(`${fixturePath}/dep-graph.json`);
+
+      const result = await inspect('.', pathToBuildConfig);
+      const actualDepGraph = result.dependencyGraph?.toJSON();
+
+      // createDepGraph(actualDepGraph, fixturePath);
```
gitphill, line 24: This is mildly dangerous, basing the expected on the result of running the test, it is painful to create dep-graphs by hand though, especially with some of these fixture sizes.

jest.config.ts
```diff
@@ -1,19 +0,0 @@
...
-    coverageReporters: ['text-summary', 'html'],
-    testMatch: ['<rootDir>/test/**/*.spec.ts', '<rootDir>/test/**/*.test.ts'],
-    globals: {
-      'ts-jest': {
-        tsconfig: '<rootDir>/test/manual/tsconfig.json',
-      },
-    },
-    modulePathIgnorePatterns: ['<rootDir>/test/.*fixtures'],
-  };
-};
```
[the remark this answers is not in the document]
Jdunsby: I thought the js one was the one coming from master 😭
I'll revert.

test/system/fixtures-with-wrappers.test.ts
gitphill: I don't know of a better way - just mentioning it in passing. Other tools do this as well (jest snapshots). We just need to make sure the generated fixture really is what we want to assert

test/system/fixtures-with-wrappers.test.ts
```diff
@@ -0,0 +1,27 @@
...
+
+describe('inspect() fixtures', () => {
+  fixtures.forEach((fixtureName) => {
+    const isKotlin = fixtureName.startsWith('kts');
+    const dsl = isKotlin ? 'Kotlin' : 'Groovy';
+    test(`${dsl} fixture: ${fixtureName}`, async () => {
+      const fixturePath = getPathToFixture(fixtureName);
+      const buildFileName = isKotlin ? 'build.gradle.kts' : 'build.gradle';
+      const pathToBuildConfig = path.join(fixturePath, buildFileName);
+      const expectedDepGraph = require(`${fixturePath}/dep-graph.json`);
```
gitphill, line 19: Better to read the JSON in and create a dep-graph object (createFromJSON).

This allows us to use actual.equals(expected) - using the dep-graph libs equality function, that doesn't care about things like order, or JSON formatting.

Also `require`ing json files can be a bit weird, because require can cache objects unexpectedly.

Jdunsby: The thing I'm not a fan of is that I'll have to assert using toBeTruthy() which means that I don't see a useful diff on a failed test, but I think not having to rely on the order/format is worth it.

gitphill: Have a look at the custom jest matcher we wrote in maven deps resolver to show dep graph equals diffs

gitphill (approved):

PR #217, opened 2022-07-20
fix: handle dependencies without lock state in strict mode

magdziarek:
- [X] Documentation written
- [X] Commit history is tidy

### What this does

Handles errors produced when trying to resolve configurations that aren't resolved in the lock file when strict mode is on.
Adds a fixture with strict lock mode on.


lib/init.gradle
```diff
@@ -171,11 +171,18 @@ def findProjectConfigs(proj, confNameFilter, confAttrSpec) {
...
+    // See Gradle docs: https://docs.gradle.org/current/userguide/declaring_dependencies.html#sec:resolvable-consumable-configs
+    def resolvable = []
+    matching.each({ it ->
+        if (!it.canBeResolved) { return }
+        try {
+            // Try accessing resolvedConfiguration to filter out configs that may cause issues in strict lock mode  
+            it.resolvedConfiguration
+            resolvable.add(it)
+        } catch (Exception ex) {
+            // Swallow the error
```
gitphill, line 183: debugLog?

gitphill (approved):

PR #221, opened 2022-08-31
fix: matching configuration error on version-catalog

magdziarek:
- [X] Documentation written
- [X] Commit history is tidy

### What this does

It moves the check for matching configurations existing in the project earlier in the script.
It changes the logic behind which projects should fail after trying to resolve configurations. 

Current behaviour: if the project has configs but none of them can be resolved, the run will fail. 
Modified behaviour: if in the project there are resolvable configs (those with the property canBeResolved set to true) that match the configuration matching filter and resolving them did not succeed, throw an error.

[Resolvable consumable configurations in Gradle docs ](https://docs.gradle.org/current/userguide/declaring_dependencies.html#sec:resolvable-consumable-configs)

Example project that fails with current behaviour: https://github.com/zone84-examples/shared-version-catalog-demo
Might also help this issue: https://github.com/snyk/cli/issues/2305


gitphill (commented):
can we add a test fixture that asserts expected behaviour?

test/fixtures-with-wrappers/kts-version-catalog-module/build.gradle.kts
```diff
@@ -0,0 +1,35 @@
+plugins {
+    `version-catalog`
+    `maven-publish`
+}
+
+group = "tech.zone84.examples.sharedcatalog"
```
gitphill, line 6: change this?

gitphill (approved):

PR #222, opened 2022-09-02
fix: subproject with same name as root

magdziarek:
- [X] Documentation written
- [X] Commit history is tidy

### What this does

Changes the way subprojects are filtered, instead of checking if the subproject has the same name as the root project, check if it has the same path. This helps with when a project contains a nested subproject with the same name as root. Do not override one with the other.

### DIFF
--- a/lib/init.gradle
+++ b/lib/init.gradle
@@ -246,7 +246,7 @@ allprojects { currProj ->
             def allSubProjectNames = []
             def seenSubprojects = [:]
             allprojects
-                .findAll({ it.name != defaultProjectName })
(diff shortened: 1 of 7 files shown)


lib/init.gradle
```diff
@@ -246,7 +246,7 @@ allprojects { currProj ->
             def allSubProjectNames = []
             def seenSubprojects = [:]
             allprojects
-                .findAll({ it.name != defaultProjectName })
+                .findAll({ it.path != task.project.path })
```
gitphill, line 249: nice! i was worried it was more complex than this

gitphill (approved):

PR #226, opened 2022-09-21
feat: normalize deps

jk05:
- [X] Documentation written
- [X] Commit history is tidy

### What this does

To support scenarios where the Gradle coordinate needs to be resolved
using a sha1 hash of a Java archive file.

Query snyk.io for the coordinate using sha1 hash.

Implement a new init.gradle task to create a sha1 map.

Only used when CLI flag `--gradle-normalize-deps` is provided.


gitphill (approved):

.circleci/config.yml
```diff
@@ -21,6 +21,7 @@ windows_defaults: &windows_defaults
     npm_config_loglevel: silent
   executor:
     name: win/default
+    size: large
```
jk05, line 24: :trollface:

PR #232, opened 2022-09-27
feat: use best practice /packages route for sha1 hash search

jk05:
- [ ] Documentation written
- [x] Commit history is tidy

### What this does

- https://github.com/snyk/maven-deps/pull/1036 introduced Snyk REST API best practices and `/rest/packages` endpoint. This PR adjusts `getMavenPackageInfo` to account for this


lib/index.ts
```diff
@@ -47,6 +50,9 @@ const cannotResolveVariantMarkers = [
   'Unable to find a matching variant of project',
 ];
 
+const PACKAGE_SEARCH_TYPE = 'maven';
+const PACKAGE_SEARCH_VERSION = '2022-09-21~beta';
```
jk05, line 54: stuck these at the top of the file so they're easy to spot. Better in a constants/common file?

lib/index.ts
```diff
@@ -500,21 +506,40 @@ export async function getMavenPackageInfo(
...
   const { res, body } = await snykHttpClient({
     method: 'get',
-    path: `/maven/coordinates/sha1/${sha1}`,
-    qs: depCoords,
+    path: `/rest/packages`,
+    qs: {
+      type: PACKAGE_SEARCH_TYPE,
+      version: PACKAGE_SEARCH_VERSION,
+      sha1,
+      /* eslint-disable @typescript-eslint/camelcase */
```
jk05, line 514: This was mentioned before by people working with openapi spec stuff, What shall we do here? I just ignored between the lines

lib/index.ts
```diff
@@ -500,21 +506,40 @@ export async function getMavenPackageInfo(
...
-    version = depCoords.version || 'unknown',
-  } = body.coordinate || {};
-  return `${groupId}:${artifactId}@${version}`;
+
+  const purlString = (body as GetPackageData)?.data[0]?.id;
+  let groupId: string;
+  let artifactId: string;
+  let version: string;
+  if (purlString) {
+    const pkg = PackageURL.fromString(purlString);
```
jk05, line 533: this will throw if `!purlString`. Ok to let it do that? Open to ideas on changes here.

lib/index.ts
```diff
@@ -500,21 +506,40 @@ export async function getMavenPackageInfo(
...
+    const { namespace, name, version: pkgVersion } = pkg;
+    groupId = namespace;
+    artifactId = name;
+    version = pkgVersion;
+  }
+
+  return `${groupId || 'unknown'}:${artifactId || 'unknown'}@${
```
jk05, line 540: `PackageURL` can return `null`, hence defaulting to `unknown` at this point and not earlier

lib/index.ts
jk05: or maybe just leave close to call site or in line?

lib/index.ts
gitphill: we should mark as unknown and continue to resolve the rest of the dep-graph, breaking the entire test is not what we want I don't think

lib/index.ts
gitphill: should we move to a search.ts file? things are starting to get a little lost in index.ts

test/functional/utilFuncs.spec.ts
```diff
@@ -32,61 +32,75 @@ describe('should split coordinate strings', () => {
...
+            data: [
+              {
+                id: 'corruptedId',
+                type: 'package',
+              },
+            ],
+          },
+        }),
+      ),
+    ).rejects.toThrow('purl is missing the required "pkg" scheme component.');
```
gitphill, line 104: i think it should return unknown and continue, instead of throwing

lib/index.ts
gitphill: that's ok

lib/index.ts
jk05: We can do more of a refactor. Ill mover some stuff around and see

gitphill (approved):

PR #235, opened 2022-10-04
fix: don't fail if matching config doesn't exist

### LINKED ISSUE #229: snyk-gradle-plugin 3.22.1 breaks --sub-project in some scenarios
Running `snyk test --sub-project=X --configuration-matching='^Y$'`, if configuration `Y` exists in subproject `X` but not other subprojects, errors out with `Matching configurations ^Y$ were not found`. 

I believe this was introduced in https://github.com/snyk/snyk-gradle-plugin/commit/79a8a084d140f76bcf8c1c73942a8670e0994627. There's a codepath under `task snykResolvedDepsJson` which invokes `findMatchingConfigs` on all subprojects even if a `--sub-project` filter is specified. And if the filter fails to match any configurations in _any_ subproject, the exception is thrown.

An easy way to reproduce this: create a standard Android project with the app subproject, add a pure java subproject to it - which will lack Android-specific configurations - and run

```
snyk  test --sub-project=app --configuration-matching='^releaseRuntimeClasspath$' --configuration-attributes='buildtype:release,usage:java-runtime'
```

magdziarek:
- [X] Documentation written
- [X] Commit history is tidy

### What this does
fix #229 
Don't throw error if requested config (passed with --configuration-matching) doesn't exist on a project .

### DIFF
--- a/lib/init.gradle
+++ b/lib/init.gradle
@@ -212,9 +212,6 @@ def matchesAttributeFilter(conf, confAttrSpec) {
 
 def findMatchingConfigs(confs, confNameFilter, confAttrSpec) {
     def matching = confs.findAll({ it.name =~ confNameFilter })
-    if (matching.isEmpty() && !confs.isEmpty()) {
-        throw new RuntimeException('Matching configurations ' + confNameFilter + ' were not found')  
-    }
     if (confAttrSpec == null) {
         // We don't have an attribute spec to match
         return matching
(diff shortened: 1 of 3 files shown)


lib/init.gradle
```diff
@@ -474,7 +471,7 @@ allprojects { Project currProj ->
                         debugLog("first level module dependencies for config `$config.name': $resConf.firstLevelModuleDependencies")
                     }
                 })
-                if (resolvedConfigs.isEmpty() && !resolvableConfigs.isEmpty()) {
+                if (!resolvableConfigs.isEmpty() && resolvedConfigs.isEmpty()) {
```
magdziarek, line 474: reads better in this order: this error should be thrown if resolvable configurations were found but for some reason were not resolved

Jdunsby (approved):

PR #237, opened 2022-10-06
test: assert gradleProjectName

magdziarek:
### What this does
Add asserts for gradleProjectName in tests


gitphill (approved):
good move, will help us detect if project identity changes

PR #242, opened 2022-11-18
fix: get lenient config if only some deps fail

magdziarek:
- [X] Documentation written
- [X] Commit history is tidy

### What this does
When a configuration cannot be fully resolved we don't resolve any of its dependencies, often returning an empty graph. It's better to return a partial result than none, like Gradle would. For this, instead of only using fully [Resolved Configuration](https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/ResolvedConfiguration.html), in case of an error, we can log the failed dependencies and use [Lenient Configuration](https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/LenientConfiguration.html) instead


lib/init.gradle
```diff
@@ -336,22 +336,28 @@ allprojects { Project currProj ->
                 def resolvableConfigs = findProjectConfigs(proj, confNameFilter, confAttrSpec)
                 def resolvedConfigs = []
                 resolvableConfigs.each({ config ->
-                    def resConf = config.resolvedConfiguration
+                    ResolvedConfiguration resConf = config.getResolvedConfiguration()
```
alexandru-dragomir, line 339: Is it possible to just use lenient configuration everywhere?

lib/init.gradle
```diff
@@ -336,22 +336,28 @@ allprojects { Project currProj ->
...
                     debugLog("config `$config.name' resolution has errors: ${resConf.hasError()}")
                     if (!resConf.hasError()) {
                         resolvedConfigs.add(resConf)
-                        debugLog("first level module dependencies for config `$config.name': $resConf.firstLevelModuleDependencies")
+                        debugLog("First level module dependencies for config `$config.name': $resConf.firstLevelModuleDependencies")
+                    } else { 
+                        LenientConfiguration lenientConf = resConf.getLenientConfiguration()
+                        debugLog("Couldn't resolve the following dependencies: ${lenientConf.getUnresolvedModuleDependencies()}")
```
alexandru-dragomir, line 346: Nitpick: to mirror the message bellow maybe "Could not resolve the following for config `$configName: ${lenientConf,getUnresolvedModuleDependencies()}"

lib/init.gradle
magdziarek: Good point - double checking now but in the end it results in the same so I don't see why not

lib/init.gradle
```diff
@@ -455,20 +461,26 @@ allprojects { Project currProj ->
                 def resolvableConfigs = findProjectConfigs(proj, confNameFilter, confAttrSpec)
                 def resolvedConfigs = []
                 resolvableConfigs.each({ config ->
```
alexandru-dragomir, line 463: can this be extracted into a common function that returns an array of resolved configurations?

lib/init.gradle
magdziarek: Tests fail for Kotlin if it's only lenient config...We can look into next? Can we merge the way it is since it's a fix for a bug?

alexandru-dragomir: From my point of view sure, but maybe leave a comment for the next person trying to simplify this? :)

alexandru-dragomir (approved):

lib/init.gradle
magdziarek: OK, will try Stepsize for this! :)

PR #257, opened 2023-03-24
fix: bug in the CLI where envVars were not being restored correctly [HEAD-218]

j-luong:
- [x] Commit history is tidy

### What this does
Fix a bug in the CLI where environment variables were not being restored correctly, due to an side effect when accessing shared memory.


j-luong:
There seems to be a issue with the CICD checks, the [latest checks](https://app.circleci.com/pipelines/github/snyk/snyk-gradle-plugin?branch=fix%2Fcopy_env_vars_before_change) in CircleCI all pass

orsagie (approved):

PR #266, opened 2023-11-15
feat: [SUP-2164] Change dependency graph root name logic

dotkas:
### What this does

There is a discrepancy between what we call our projects and what we call our root nodes in the dependency graphs. This sounds like it should not matter, but it does.

We have a regression introduced after a bugfix was made last month, where `snyk test` now does not correctly pick up ignores in the ui after a `snyk monitor` on certain types of Gradle projects.

Doing this change will fix that.

### Problem

(Links to internal projects)

* [This](https://github.com/snyk/registry/blob/ac4373282fbc6d6a145cc9bf4fab2518e0d7b144/src/web/routes/api/v1/monitor.ts#L165-L168) is how we make a project id in `monitor`.
* [This](https://github.com/snyk/registry/blob/d078339f187dfe5b9de4789e03fa823e54bac81f/src/web/routes/api/v1/test/dep-graph.ts#L125-L130) is how we make a project id in `test`.

Since the bugfix last month, root package names no longer match the project names in Gradle projects. Doing a `--list-deps` or looking at the UI will show one name in the project name and another in the root package list.

This is the root cause of the bug why `snyk test` does not pick up changes, as the project identity that is being tested against is generated from the root package name, not the "defaultProjectName" like `snyk monitor`.

### Breaking change

I'm bumping the major version for this one, unless anybody objects. Not **technically** a breaking change, as this should not cause orphaned projects. It should merely create a new element in the history i the same project - but out of caution i find it makes sense, just to show that we're changing expected behavior. In case any consumer is relying on this value.

## Project identity?

The `projectName` should not change because of these changes. This change actually more aligns the root package name to what is already done for project names: https://github.com/snyk/snyk-gradle-plugin/blob/a444194f1199afece07232f4568773f36f9bfb9f/lib/index.ts#L306-L316

## Snapshot churn?

This should change dependency graphs for CLI monitor projects, and only a part of Gradle customers, so after discussing with some stakeholders in Team Fix, have deemed it should not be dangerous to snapshot churn.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -6,19 +6,14 @@ import * as tmp from 'tmp';
 import * as pMap from 'p-map';
 import * as chalk from 'chalk';
 import { DepGraph } from '@snyk/dep-graph';
-import debugModule = require('debug');
 import { legacyCommon, legacyPlugin as api } from '@snyk/cli-interface';
 
 import { MissingSubProjectError } from './errors';
 import { getGradleAttributesPretty } from './gradle-attributes-pretty';
 import { buildGraph, SnykGraph } from './graph';
-import type {
-  CoordinateMap,
-  PomCoords,
-  Sha1Map,
-  SnykHttpClient,
-} from './types';
+import type { CoordinateMap, PomCoords, Sha1Map, SnykHttpClient, } from './types';
 import { getMavenPackageInfo } from './search';
+import debugModule = require('debug');
 
 type ScannedProject = legacyCommon.ScannedProject;
 
@@ -98,12 +93,12 @@ export async function inspect(
 ): Promise<api.InspectResult> {
   debugLog(
     'Gradle inspect called with: ' +
-      JSON.stringify({
-        root,
-        targetFile,
-        allSubProjects: (options as any)?.allSubProjects,
-        subProject: (options as any)?.subProject,
-      }),
+    JSON.stringify({
+      root,
+      targetFile,
+      allSubProjects: (options as any)?.allSubProjects,
+      subProject: (options as any)?.subProject,
+    }),
   );
 
   if (!options) {
@@ -216,13 +211,13 @@ function extractJsonFromScriptOutput(stdoutText: string): JsonDepsScriptResult {
   if (jsonLine === null) {
     throw new Error(
       'No line prefixed with "JSONDEPS " was returned; full output:\n' +
-        stdoutText,
+      stdoutText,
     );
   }
   debugLog(
     'The command produced JSONDEPS output of ' +
-      jsonLine!.length +
-      ' characters',
+    jsonLine!.length +
+    ' characters',
   );
   return JSON.parse(jsonLine!);
 }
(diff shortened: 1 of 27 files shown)


lib/index.ts
```diff
@@ -626,16 +621,11 @@ export async function processProjectsInExtractedJSON(
...
     const isSubProject = projectId !== defaultProjectKey;
 
-    let projectName = isValidRootDir ? path.basename(root) : defaultProject;
-
+    let projectName = defaultProject;
     if (isSubProject) {
-      projectName = isValidRootDir
-        ? `${path.basename(root)}/${projectId}`
-        : `${defaultProject}/${projectId}`;
+      projectName = `${defaultProject}/${projectId}`;
```
dotkas, line 628: Should also make it consistent with https://github.com/snyk/snyk-gradle-plugin/blob/a444194f1199afece07232f4568773f36f9bfb9f/lib/index.ts#L306-L309

dotkas: Note that this is not the actual `projectName` that is being changed, despite the variable name. It's the root package name only. 

Actual `projectName` stays as it is.

test/manual/gradle-stdout.spec.ts
```diff
@@ -5,14 +5,14 @@ describe('findProjectsInExtractedJSON', () => {
...
-    rootDir        | targetFile
-    ${''}          | ${'build.gradle'}
-    ${null}        | ${'build.gradle'}
-    ${undefined}   | ${'build.gradle'}
-    ${fakeRootDir} | ${path.join(fakeRootDir, 'build.gradle')}
+    targetFile
+    ${'build.gradle'}
+    ${'build.gradle'}
+    ${'build.gradle'}
+    ${path.join(fakeRootDir, 'build.gradle')}
```
dotkas, line 12: `rootDir` was not needed as we're no longer using it for `rootPkg` name generation.

lib/index.ts
dotkas: If we accept it we should probably unify it in a function.

test/fixtures-with-wrappers/with-lock-file/dep-graph.json
dotkas, line 1: Honest answer, I am unsure why this file is changed. It was failing on locally, but also on CircleCI, so I ssh'd in there and tried to look, to ensure my local setup was not the culprit, and CircleCI showed the same changes.

dotkas: Seems the pipeline has been failing for a long time, so most likely unrelated to this change.

JamesPatrickGill (approved):

gitphill (approved):

anthogez (approved):

lib/index.ts
danielroymoore: Should the variable be named to something other than projectName to avoid future confusion?

danielroymoore (approved):

PR #268, opened 2023-11-22
Use complete project path when filtering subprojects

### LINKED ISSUE #195: Please use complete sub project path instead of name to check for onlyProj
https://github.com/snyk/snyk-gradle-plugin/blob/9989b21137695bedc5182f551e6d3299e6eebb23/lib/init.gradle#L248

Gradle multiproject setup can be nested.
One can setup the following structure:
```
settings.gradle
build.gradle
sub1/foo/build.gradle
sub2/foo/build.gradle
```

The `settings.gradle` could look like this:
```groovy
rootProjectName="myProject"
include ':sub1:foo'
include ':sub2:foo'
```

So here we have *four* subprojects!
* `:sub1`
* `:sub1:foo`
* `:sub2`
* `:sub2:foo`

When declaring the `--sub-project` one has to pass the sub project name without the path (eg `--sub-project=foo`). It would be cleaner and actually more correct to use the full project path `--sub-project=:sub1:foo`.
Gradle exposes this in the `Project` class.
> Projects are arranged into a hierarchy of projects. A project has a name, and a fully qualified path which uniquely identifies it in the hierarchy.

see: https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#getPath--

damianw:
- [ ] Tests written and linted
- [X] Documentation written
- [X] Commit history is tidy

### What this does

This resolves #195, allowing the tool to properly recognize `--sub-project` arguments that refer to nested subprojects, e.g. `foo/bar` for the Gradle project path `:foo:bar`. Currently, this does not work.

### Notes for the reviewer

I'm not sure how to write the test for this, since I haven't dug into the repo too much. Please let me know if you have a recommendation for anything else that needs to be done.

Ran with `--sub-project` for a nested Gradle subproject, e.g. `foo/bar`, and found it to be recognized.

### DIFF
--- a/lib/init.gradle
+++ b/lib/init.gradle
@@ -334,7 +334,7 @@ allprojects { Project currProj ->
             def shouldScanProject = {
                 onlyProj == null ||
                 (onlyProj == '.' && it.name == defaultProjectName) ||
-                it.name == onlyProj
+                formatPath(it.path) == onlyProj
             }
             def projectsDict = [:]
 


anthogez:
Thanks @damianw for your collaboration, we will review your changes and loop back to you.

damianw:
Thanks @37IulianPopovici! I don't have permission to create a branch on this repo, if that's what you're asking. I'm happy to do that if given permission, but I also don't mind if you commandeer this change and take ownership of it.

37IulianPopovici:
Hey @damianw, thanks a lot for your patience! Waiting for the [cli PR](https://github.com/snyk/cli/pull/5004) to be approved and released

damianw:
Thanks, appreciate you looking into it!

PR #280, opened 2024-01-15
feat: [OSM-892] added support for gradle 8

37IulianPopovici:
- [X] Documentation written
- [X] Commit history is tidy

### What this does

Adds support for Gradle 8
Enhanced tests to also run on Gradle 7 and Gradle 8


lib/init.gradle
```diff
@@ -348,6 +349,14 @@ allprojects { Project currProj ->
...
                 debugLog("processing project: name=$proj.name; path=$proj.path")
 
+                String aaaa = formatPath(proj.path)
+                String name = proj.name
+                String pathh = proj.path
+                println("proj.name $name")
+                println("proj.path $pathh")
+                println("onlyProj $onlyProj")
+                println("formatted path $aaaa")
+
```
dotkas, line 359: LGTM 🤣

test/fixtures/gradle-kts/build.gradle.kts
dotkas, line 1: I might be missing something but are you replacing existing tests for Gradle 7 with new tests for Gradle 8? If so, shouldn't we support both?

test/functional/gradle-plugin.spec.ts
```diff
@@ -159,25 +160,28 @@ describe('Gradle Plugin', () => {
...
-    );
-    expect(result).toEqual(
-      expect.arrayContaining([
-        'snykResolvedDepsJson',
-        '-q',
-        '-Dorg.gradle.parallel=',
-        '-Dorg.gradle.console=plain',
-        '-PonlySubProject=.',
-        '-I',
+  it.each(NO_CONFIGURATION_CACHE_VERSIONS)(
```
dotkas, line 163: nit, but I think the constant is making it harder to read what this test does if the variable is not re-used anywhere.

test/fixtures/gradle-kts/build.gradle.kts
37IulianPopovici: No, before this PR, tests were running only on Gradle 4, 5 and 6 with Java version 8. What I've did was to modify the tests so they could also run with Java 17 (needed for Gradle 7 and Gradle 8)

dotkas: Ok, but it still works for 4,5,6 on Java 8 with these changes? 🤷🏻

dotkas (approved):

PR #289, opened 2024-09-12
feat: Update CI Window Machine

DOlufemi:
- [ ] Documentation written
- [x] Commit history is tidy

### What this does

- Upgrade Circle CI window machine version
- Ensure appropriate JDK version is used when running tests 
- Upgrade test

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -415,7 +415,7 @@ function getVersionBuildInfo(
   }
 }
 
-async function getGradleVersion(
+export async function getGradleVersion(
   root: string,
   command: string,
 ): Promise<string> {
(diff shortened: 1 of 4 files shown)


lib/index.ts
```diff
@@ -415,7 +415,7 @@ function getVersionBuildInfo(
   }
 }
 
-async function getGradleVersion(
+export async function getGradleVersion(
```
andrei-cacio, line 418: is this needed?

DOlufemi: Yes, this is now being used to set up the test environment.

andrei-cacio (approved):

PR #299, opened 2025-03-17
feat: module artifacts with classifier

gitphill:
There can be multiple module artifacts. At the moment we assume
there can only by one artifact per resolved dependency. This
is the most common arrangement, however there can be multiple
artifacts when several 'classifiers' or different 'types' are
referenced. For example if both the test and sources jar files
are required.

Loop over each module artifact and use the classifier in the
node id, always appending ':type' where 'jar' is the default
type and optionally appending ':classifier'.

In order to find vulnerabilities against the resolved dependency
the package identity remains the same,
'groupid:artifactid@version'.

The net effect here is that we now correctly identify artifacts
and create a node per artifact.

One fix that has occurred by only considering those dependencies
with an artifact is that those dependencies without artifacts
are excluded. This can happen when using 'platform' dependencies.
These are constraints that don't require dependencies but control
the versions of dependencies required either directly or
transitively. They are analogous to maven boms and
dependencyManagement.


test/fixtures-with-wrappers/platform-project-mvn-bom/dep-graph.json
```diff
@@ -11,13 +11,6 @@
         "version": "unspecified"
       }
     },
-    {
-      "id": "org.springframework.boot:spring-boot-dependencies@1.5.8.RELEASE",
```
gitphill, line 15: this is not a dependency, it is a platform constraint that controls the versions of other spring dependencies in the project.

test/fixtures-with-wrappers/version-catalog-toml/dep-graph.json
```diff
@@ -60,13 +60,6 @@
         "version": "5.8.2"
       }
     },
-    {
-      "id": "org.junit:junit-bom@5.8.2",
```
gitphill, line 64: this is not a dependency, it is a platform constraint that controls the versions of other spring dependencies in the project.

37IulianPopovici (approved):

lib/graph.ts
```diff
@@ -52,18 +53,20 @@ export async function buildGraph(
...
-    if (coordinateMap) {
-      if (coordinateMap[id]) {
-        id = coordinateMap[id];
-        const [newName, newVersion] = id.split('@');
+    if (sha1Map) {
+      if (sha1Map[id]) {
+        id = sha1Map[id];
+        const coord = parseCoordinate(id);
+        const newName = `${coord.groupId}:${coord.artifactId}`;
+        const newVersion = coord.version;
```
gitphill, line 61: This is subtle, it ensures we retain 'GAV' `groupId:artifactId@version` as the pkg identity. Snyk vulnerabilities are identified by GAV, so to match pkg to vuln we need to ignore type and classifier.

test/fixtures-with-wrappers/basic-with-deps/dep-graph-gradleNormalizeDeps-failed-search.json
```diff
@@ -1,167 +0,0 @@
...
-  "pkgs": [
-    {
-      "id": "basic-with-deps@unspecified",
-      "info": {
-        "name": "basic-with-deps",
-        "version": "unspecified"
-      }
-    },
-    {
-      "id": "unknown:guava-87e0fd1df874ea3cbe577702fe6f17068b790fd8@unknown",
```
gitphill, line 15: I don't think it was useful putting sha1 into the artifact id, unresolved sha1 hashes can be seen in the debug logs

james-snyk (approved):

Snapp949 (commented):
Ty for the help

Snapp949 (commented):
Ty for the help

PR #302, opened 2025-04-15
fix: error when path contains space on windows

james-snyk:
Fixes an issue where if a users file path contains any spaces its causes the application to error (windows only). To fix this we need to wrap all paths in " for windows users. This fixes the issues but causes another due to how the command gets executed on windows. This can be fixed by running the command in a shell to enable the " to be interpret correctly by windows.


lib/index.ts
```diff
@@ -701,7 +700,12 @@ function buildArgs(
     }
     args.push('--build-file');
 
-    args.push(resolvedTargetFilePath);
+    const isWinLocal = /^win/.test(os.platform());
+    if (isWinLocal) {
+      args.push(`"${resolvedTargetFilePath}"`);
+    } else {
+      args.push(resolvedTargetFilePath);
+    }
```
MarcusArdelean, line 708: non-blocker

I see you used roughly the same code in 2 different parts. Maybe worth putting in a standalone function in a `utils.ts` file or something like that. Just so that it can be used in the future in other places easily.

MarcusArdelean (approved):
Nice job 🔥

DOlufemi (approved):

PR #309, opened 2025-06-27
fix: Pre-compute child links and use set for ancestry

calhar-snyk:
- [ ] Tests written and linted
- [x] Documentation written
- [x] Commit history is tidy

### What this does

This provides further performance improvements to the dep-graph generation of the `snyk-gradle-plugin`.

Specifically:
1. `findChildren` was recalculating the children for each node afresh. This added additional complexity of `O(N * max_P)` to each loop.
2. Ancestry was using `Array.includes` which is `O(N)`. We still need to pay the cost when generating the next set of ancestors -- however, checking if we're in a cyclic section is now `O(1)`.

This takes us from:
```
  snyk-gradle-plugin The command produced JSONDEPS output of 117528 characters +34ms
  snyk-gradle-plugin The following attributes and their possible values were found in your configurations:            org.gradle.category: library, verification
  snyk-gradle-plugin               org.gradle.usage: java-runtime, java-api
  snyk-gradle-plugin     org.gradle.libraryelements: classes, jar
  snyk-gradle-plugin org.gradle.dependency.bundling: external
  snyk-gradle-plugin     org.gradle.jvm.environment: standard-jvm
  snyk-gradle-plugin         org.gradle.jvm.version: 17
  snyk-gradle-plugin    org.gradle.verificationtype: test-results, main-sources
  snyk-gradle-plugin      org.gradle.testsuite.name: test +1ms
  snyk-test Found 1 projects from 1 detected manifests +48s
```

To:
```
  snyk-gradle-plugin The command produced JSONDEPS output of 117528 characters +28ms
  snyk-gradle-plugin The following attributes and their possible values were found in your configurations:            org.gradle.category: library, verification
  snyk-gradle-plugin               org.gradle.usage: java-runtime, java-api
  snyk-gradle-plugin     org.gradle.libraryelements: classes, jar
  snyk-gradle-plugin org.gradle.dependency.bundling: external
  snyk-gradle-plugin     org.gradle.jvm.environment: standard-jvm
  snyk-gradle-plugin         org.gradle.jvm.version: 17
  snyk-gradle-plugin    org.gradle.verificationtype: test-results, main-sources
  snyk-gradle-plugin      org.gradle.testsuite.name: test +1ms
  snyk-test Found 1 projects from 1 detected manifests +11s
```

Inconveniently -- I've changed the `snyk-gradle-plugin` entry in the `package.json` for the CLI to point to my local repo and then building the CLI locally using `npm install && npm run build-cli:prod`.

I've then been testing on a project locally using this command and tracking the timing information
```
path/to/cli/bin/snyk test --all-projects --print-graph -d
```

### DIFF
--- a/lib/graph.ts
+++ b/lib/graph.ts
@@ -15,7 +15,34 @@ export interface GradleGraph {
 interface QueueItem {
   id: string;
   parentId: string;
-  ancestry: string[];
+  ancestry: Set<string>;
+}
+
+function precomputeChildrenMap(
+  gradleGraph: GradleGraph,
+): Map<string, string[]> {
+  const childrenMap = new Map<string, string[]>();
+  for (const id of Object.keys(gradleGraph)) {
+    const node = gradleGraph[id];
+    if (node?.parentIds) {
+      for (const parentId of node.parentIds) {
+        if (!childrenMap.has(parentId)) {
+          childrenMap.set(parentId, []);
+        }
+        childrenMap.get(parentId).push(id);
+      }
+    }
+  }
+  return childrenMap;
+}
+
+function findChildren(
+  parentId: string,
+  ancestry: Set<string>,
+  childrenMap: Map<string, string[]>,
+): QueueItem[] {
+  const childrenIds = childrenMap.get(parentId) || [];
+  return childrenIds.map((id) => ({ id, ancestry, parentId }));
 }
 
 export async function buildGraph(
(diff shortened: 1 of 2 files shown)


lib/graph.ts
```diff
@@ -14,7 +15,39 @@ export interface GradleGraph {
...
+}
+
+function findChildren(
+  parentId: string,
+  ancestry: Set<string>,
+  childrenMap: Map<string, string[]>,
+): QueueItem[] {
+  const result: QueueItem[] = [];
+  const childrenIds = childrenMap.get(parentId) || [];
+  for (const id of childrenIds) {
```
calhar-snyk, line 47: This could instead just be:
```
return childrenIds.map(id => ({ id, ancestry, parentId });
```

JamesPatrickGill (commented):
So am happy with the changes but want to clarify one thing.

We aren't pruning any differently than we are previously? 

It doesn't look like it but we need to make sure the outputs are the same as before and if not we need to make sure we broadcast the changes users will see.

PR #321, opened 2025-09-19
fix: filter out configs without deps

gitphill:
- [X] Documentation written
- [X] Commit history is tidy

### What this does

Before using a configuration, check that it has resolvable dependencies. Each configuraiton must have at least one firstLevelModuleDependencies that comes from a resolvable configuration.

This filters out deprecated configurations, like compile and runtime.

By allowing these configurations to be processed we incorrectly resolve dependencies for them.


gitphill:
i'm not sure we can do this, pretty sure it breaks other setups using custom configurations

PR #337, opened 2026-05-19
feat: Update init.gradle to use non-deprecated Gradle APIs

cjheppell:
### ⚠️ Major version bump required

As this fundamentally changes the resolution algorithm, this constitutes a major version bump.

### What this does

Switches the Gradle dependency walker in `lib/init.gradle` from the v5 `ResolvedConfiguration` API to the modern `ResolutionResult` + `ArtifactView` APIs. Same `JSONDEPS` schema; `lib/graph.ts` unchanged.

### Why

`ResolvedConfiguration.getFirstLevelModuleDependencies()` + `ResolvedDependency.getChildren()` returns stale per-project metadata when traversing project deps with multiple variants (e.g. `java-library` + `java-test-fixtures`). Gradle's Javadoc explicitly says: *"legacy API. Avoid this class for new code."* `ResolutionResult` applies cross-project conflict resolution correctly — same machinery as `gradle dependencyInsight`.

### Key walker changes

- **Cross-project conflict resolution**: walks `resolutionResult.root` recursively, so the version reported per project matches what Gradle actually puts on the classpath.
- **Multi-config walk**: each resolvable configuration walked separately (v5 flattened all configs into one walk). Deps appearing on multiple classpaths now produce one real node and one `:pruned` placeholder per additional arrival path.
- **Multi-parent edges**: a component with multiple artifacts (e.g. classifier variants) attaches its transitives under each artifact — preserves v5's per-classifier arrival paths.
- **No-artifact components skipped**: BOMs, Gradle platforms, project deps with no jar are no longer emitted as logical nodes. Their transitives attach to the BOM's parent instead.
- **Publication configs skipped**: Gradle's auto-created `default` / `archives` configs aren't walked. v5 walked them too but via the first-level API, which is empty for these configs — so v6 surfaces their extends-from transitives that v5 silently dropped.
- **Lock-rejection noise suppressed**: under strict locks, Gradle emits an `UnresolvedDependencyResult` for the rejected request next to the resolved one. v6 drops those — they're not on any classpath.
- **Determinism**: `ResolutionResult.dependencies` is an unordered `Set`; sort by selector display name before iterating.

### Fixture impact

Most fixtures (12/20) unchanged. The rest are **strictly additive** — every diff is more information, never less:

| Δ pkgs | Fixture | Why |
|---|---|---|
| +3 | `version-catalog-toml` | BOM transitives v5's stale-metadata walk missed |
| +1 | `init-gradle/app` | `junit-jupiter-engine` via the junit BOM, also missed by v5 |
| +1 | `basic-with-failed-dep` | declared-but-failed dep (`axis:axis@badVersion-7.7`) — v5 silently dropped |

Plus 5 fixtures gain `:pruned` placeholders (no new packages — multi-config walk surfaces arrival paths v5 flattened away).

`nodeId` format changed from `name@version` to `name:type@version` (always includes the artifact type). Doesn't affect vuln scanning — only the `pkgs` array is consumed downstream, and across all fixtures the net change is **+5 packages gained, 0 lost**.


lib/init.gradle
```diff
@@ -121,94 +121,249 @@ def hashString(String str) {
...
+}
+
+def getGradleGraphWithSha1Map(Iterable resolvableConfigs, Sha1Map sha1Map) {
+    if (!sha1Map) sha1Map = new Sha1Map()
+    def graph = new GradleGraph('root-node')
+    walkResolvableConfigs(resolvableConfigs, graph, sha1Map, true)
+    return [graph: graph.nodes, sha1Map: sha1Map]
+}
+
+def walkResolvableConfigs(Iterable resolvableConfigs, GradleGraph graph, Sha1Map sha1Map, boolean withSha1Map) {
```
calhar-snyk: Nitpick: `withSha1Map` is implicit if `sha1Map` is non-null

test/fixtures-with-wrappers/version-catalog-settings/dep-graph.json
```diff
@@ -12,16 +12,16 @@
       }
     },
     {
-      "id": "org.codehaus.groovy:groovy@3.0.5",
+      "id": "org.codehaus.groovy:groovy-json@3.0.5",
```
calhar-snyk, line 15: There's a fair amount of order shuffling in the output.
Probably not a problem, but do we know why?

cjheppell: My understanding is that the the way that we now traverse the various dependencies is different to the old API, which leads to a different order in how we build up the dep graph.

I added some code to ensure that things like hash-iteration order doesn't catch us out and so that we're deterministic _between invocations_. See: https://github.com/snyk/snyk-gradle-plugin/pull/337/changes/BASE..6c2008413093cba44412250337e4d8789269e4ae#diff-e1388fa57aa98ffb983e5870055574d6a3434baa4fcd99d56e0f931e21d4607eR274 

But the update to the non-deprecated API fundamentally changes the way in which we traverse dependencies, so changes the graph ordering we build. We _could_ put some effort in to preserve the existing ordering. But I'm not sure its worth the complexity.

lib/init.gradle
cjheppell: Good catch. 🤦

test/fixtures-with-wrappers/basic-with-failed-dep/dep-graph.json
```diff
@@ -11,13 +11,34 @@
         "version": "unspecified"
       }
     },
+    {
+      "id": "axis:axis@badVersion-7.7",
```
gitphill, line 15: this one is interesting; is this not actually failing now? in which case the test fixture may need updating (name and README)

test/fixtures-with-wrappers/init-gradle/app/dep-graph.json
```diff
@@ -106,9 +113,6 @@
           {
             "nodeId": "org.apache.commons:commons-text:jar@1.12.0"
           },
-          {
-            "nodeId": "init-gradle:utilities:java-classes-directory@unspecified"
```
gitphill, line 110: here we're not representing internal sub projects as nodes anymore?

test/fixtures-with-wrappers/basic-with-failed-dep/dep-graph.json
cjheppell: So, this _is_ a behavioural change. We can revert this if preferred.

Resolution fails (because this doesn't exist) but we now include it in the dep graph. I think @calhar-snyk was of the opinion this is preferable, but I'm happy to change if we don't like it.

test/fixtures-with-wrappers/init-gradle/app/dep-graph.json
cjheppell: It's more that artifact variants of this nature (java-classes-directory) weren't being properly identified. We were falling back to `:jar` so it looked like these were being removed. This looks like a regression.

That's a symptom of the new API not exposing the same information as the old one in certain Gradle versions.

I've reinstated this for Gradle 5.1+ upwards, but unfortunately we aren't able to get the same information for <5.1 as the api we need `getVariant()` wasn't introduced until then.

See: https://github.com/snyk/snyk-gradle-plugin/pull/337/commits/25910ca2f0739b4e68cd105b84b740781feb590d

calhar-snyk (approved):

PR #345, opened 2026-07-17
feat: add includeComponentMetadata support to Gradle resolver

calhar-snyk:
## What

Adds opt-in **component-metadata** node labels to the Gradle resolver, using the
same cross-ecosystem vocabulary as `snyk-mvn-plugin` and `nodejs-lockfile-parser`:

- `hash:<alg>` — lowercase-hex `md5` / `sha-1` / `sha-256` / `sha-512` digests of
  the resolved artifact file.
- `distribution:url` — the real remote URL Gradle read the artifact from.

Enabled via the `includeComponentMetadata` inspect option (passed to the init
script as `-PsnykIncludeComponentMetadata=true`). Off by default; no behaviour
change unless requested.

`distribution:url` is **true provenance only** — Gradle's public API exposes no
per-artifact source repository, so the real resource-read URLs are captured via
an internal build-operation listener. That path is best-effort: if the internal
API is unavailable on a given Gradle version, or nothing is re-fetched on a warm
cache, the node simply carries no `distribution:url` (never a reconstructed
guess). `gradleRefreshDependencies` forces `--refresh-dependencies` so the reads
fire and `distribution:url` is populated even on a warm cache.

## On `distribution:url` sanitising

The value is reduced at capture to `scheme://host[:port]/path`. The realistic
in-URL credential vector is **userInfo** — a repo configured as
`https://user:pass@host/repo` carries it on every artifact URL — which this
strips so it can't reach the dep-graph/SBOM. Query and fragment are dropped as
**canonicalisation**: they are not part of the artifact's durable location (at
most a request-scoped redirect artefact), so the bare path is the correct,
reproducible provenance value. It is not a defence against exotic signed-URL
token schemes (Azure SAS / S3-presigned) — those don't arise for normal
header-authenticated, path-based Maven/Gradle repos. `canonicalUrl` fails closed:
an un-sanitisable URL is dropped rather than risking a leak.

## Testing

- `test/system/component-metadata.test.ts`: flag on → every resolved external
  node carries all four hashes as lowercase hex; flag off → no such labels.
  Passing locally on Gradle 9.4.1 / JDK 21.

### Example `snyk sbom` output with built cli
```
  "components": [
    {
      "bom-ref": "2-org.jetbrains.kotlin:kotlin-stdlib-jdk8@2.2.0",
      "type": "library",
      "group": "org.jetbrains.kotlin",
      "name": "org.jetbrains.kotlin:kotlin-stdlib-jdk8",
      "version": "2.2.0",
      "hashes": [
        {
          "alg": "MD5",
          "content": "5fe4006cd33776d7709096d3f9fa97ca"
        },
        {
          "alg": "SHA-1",
          "content": "ed572b04cd66acce7837ad753e4fe64d9c536025"
        },
        {
          "alg": "SHA-256",
          "content": "adc16648dbbcf35b0d10e7ec301c35d746d1c2fe460c606aba59f12b117cf9b0"
        },
        {
          "alg": "SHA-512",
          "content": "6307f0854f21811b0b5ff54ada950d44391a813fa41bf3c0de74ebee4d5ca6b861eaa1bec9e6aa991c7942ea0b67a6504adb8198dd5a42d86240d58465e29dfb"
        }
      ],
      "licenses": [
        {
          "expression": "Apache-2.0"
        }
      ],
      "purl": "pkg:maven/org.jetbrains.kotlin/kotlin-stdlib-jdk8@2.2.0",
      "externalReferences": [
        {
          "url": "https://repo.maven.apache.org/maven2/org/jetbrains/kotlin/kotlin-stdlib-jdk8/2.2.0/kotlin-stdlib-jdk8-2.2.0.jar",
          "type": "distribution"
        }
      ]
    },```

## Notes

- The init-script (Groovy) logic is only exercised by the system suite
  (`npm run test:system`), not the fast lint/functional lane.
- Companion change in `cli-extension-dep-graph` wires the same feature through
  the native resolver; the `digestFile` and `canonicalUrl` helpers are identical
  across both.

Draft: opening for review of the approach before finalising.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -671,6 +671,17 @@ function buildArgs(
     args.push(`-PconfAttr=${options['configuration-attributes']}`);
   }
 
+  // Opt-in component-metadata labels (hash:<alg>, distribution:url). The init
+  // script only does the extra work when this property is present.
+  if (options.includeComponentMetadata) {
+    args.push('-PsnykIncludeComponentMetadata=true');
+    // Forces network re-validation of metadata so distribution:url can be
+    // captured on a warm cache. Only meaningful alongside the metadata flag.
+    if (options.gradleRefreshDependencies) {
+      args.push('--refresh-dependencies');
+    }
+  }
+
   if (options.initScript) {
     const formattedInitScript = path.resolve(options.initScript);
     args.push('--init-script', formattedInitScript);
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -687,8 +687,13 @@ function buildArgs(
     args.push('--init-script', formattedInitScript);
   }
 
-  const isWin = /^win/.test(os.platform());
-  if (isWin && !options.daemon) {
+  // Default to a fresh, one-shot JVM unless a daemon is explicitly requested.
+  // On Windows a daemon otherwise leaves the process hanging from Node's
+  // standpoint; on Unix a reused daemon accumulates heap across repeated
+  // invocations (e.g. a full test suite run), which under a memory-constrained
+  // runner tips it into an OOM kill. A single one-shot scan gains nothing from
+  // the daemon, so opt out on every platform.
+  if (!options.daemon) {
     args.push('--no-daemon');
   }
 
--- a/lib/init.gradle
+++ b/lib/init.gradle
@@ -126,15 +126,28 @@ class Sha1Map {
 // `algorithms` maps the JDK algorithm name to its output label; returns
 // {label: lowercase-hex}. Shared by every file-hashing helper below.
 def digestFile(File file, Map<String, String> algorithms) {
-    def digests = algorithms.collectEntries { javaName, _label ->
-        [(javaName): java.security.MessageDigest.getInstance(javaName)]
+    def javaNames = algorithms.keySet().toList()
+    def mds = javaNames.collect { java.security.MessageDigest.getInstance(it) }
+    // Reuse a single 64 KB buffer for the whole file. Groovy's
+    // File.eachByte(int, Closure) allocates a fresh byte[] copy per chunk, so
+    // hashing large artifacts churns enough short-lived garbage to inflate the
+    // Gradle daemon's peak heap; a manual read loop keeps allocation ~constant.
+    // Digests are identical either way — MessageDigest is chunking-invariant.
+    // All state is local, so this stays safe under parallel per-project tasks.
+    byte[] buffer = new byte[64 * 1024]
+    file.withInputStream { input ->
+        int read
+        while ((read = input.read(buffer)) != -1) {
+            for (md in mds) {
+                md.update(buffer, 0, read)
+            }
+        }
     }
-    file.eachByte(1024 * 8) { buffer, len ->
-        digests.each { _name, md -> md.update(buffer, 0, len) }
+    def result = [:]
+    javaNames.eachWithIndex { javaName, i ->
+        result[algorithms[javaName]] = mds[i].digest().encodeHex().toString()
     }
-    return digests.collectEntries { javaName, md ->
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: lib/index.ts [hunk 1]
```diff
@@ -671,13 +671,29 @@ function buildArgs(
...
 
+  // Opt-in component-metadata labels (hash:<alg>, distribution:url). The init
+  // script only does the extra work when this property is present.
+  if (options.includeComponentMetadata) {
+    args.push('-PsnykIncludeComponentMetadata=true');
+    // Forces network re-validation of metadata so distribution:url can be
+    // captured on a warm cache. Only meaningful alongside the metadata flag.
+    if (options.gradleRefreshDependencies) {
+      args.push('--refresh-dependencies');
+    }
```
cjheppell, line 682: this means its a no-op without the above flag?
would it make more sense to error out if provided without?

lib/init.gradle
```diff
@@ -108,17 +122,218 @@ class Sha1Map {
...
+        // deems registry-based rather than server-based — notably hosts
+        // containing an underscore, common on internal Maven mirrors — which
+        // would silently drop the host from the label. getAuthority() preserves
+        // it, so we strip the "user:pass@" prefix ourselves to keep credentials
+        // out of the dep-graph/SBOM.
+        def authority = u.authority
+        if (authority == null) return null
+        int at = authority.lastIndexOf('@')
+        if (at >= 0) authority = authority.substring(at + 1)
+        return new URI(u.scheme, authority, u.path, null, null).toString()
```
cjheppell: in other places we've constructed a new URI and explicitly set the username/password to null or empty string. Can we do the same here?
I worry about `@` values in usernames/passwords causing trouble here.

lib/index.ts
calhar-snyk: Maybe? I think a no-op is fine here. I think flags should error when they're given invalid input -- but in this case, the flag just isn't particularly meaningful outside of this usecase. I suppose an argument could be made for supporting it as a general flag -- a dependency scan with refreshed dependencies has _some_ value -- but I don't think it's going to be a common usecase.

lib/init.gradle
calhar-snyk: Your concern here is valid. I'll adjust it

calhar-snyk: `java.net.URI` implements RFC 2396, whose hostname grammar allows only alphanumerics and -. Any authority it can't parse as `host[:port]` -- one underscore is enough -- falls back to an opaque "registry name", where `getHost()`, `getPort()` and `getUserInfo()` are all left null and only `getAuthority()` survives. It fails silently, no exception. So on an internal mirror like `my_host.internal`, the component accessors give you nothing to rebuild from, and a `getUserInfo() != null` guard would skip credential stripping exactly where credentials are present.

So we have a happy path (`getHost() != null`) and a sad path where the host fails to parse and we have to handle the authority instead.

calhar-snyk: My follow on concern was whether '@' could appear in the hostname rather than in a credential combination. To which:

It can, in exactly one way, and it changes the implementation. RFC 2396's `reg_name` permits `@`, and more importantly `getAuthority()` decodes percent-escapes:

|                                | `getAuthority()`      | `getRawAuthority()`     |
| ------------------------------ | --------------------- | ----------------------- |
| `https://my%40host.internal/p` | `my@host.internal` ⚠️ | `my%40host.internal` ✅ |

On the decoded form, last-`@` would emit `host.internal` -- a truncated host. So the fallback splits `getRawAuthority()`, where escapes stay encoded and the only unencoded `@` is the real userinfo delimiter. `my%40host.internal` now passes through intact.

Note: `foo@bar` never reaches the fallback at all -- Java parses that as server-based (host=bar, userInfo=foo), so it takes the happy path.

cjheppell (approved):

Issue #344, opened 2026-07-14
Support configuration-cache/isolated-project feature


ganadist:
It looks `Isolated Projects` feature graduate `experimental`, soon and will promote as incubation stage.
https://github.com/gradle/gradle/pull/38302

snyk gradle plugin (and cli) don't support configuration-cache/isolated-project feature, yet.
https://github.com/snyk/snyk-gradle-plugin/blob/e404c211c782c82c6c386895f50e7456a7ea14cd/lib/index.ts#L704-L710

By this block, snyk cli was failed on gradle project which is enabled `Isolated Projects` feature.
```
./gradlew     
           snykResolvedDepsJson -q --project-dir /some/where/projectDir -                                
           Pconfiguration=.*([Dd]ebug|[Rr]elease)(Compile|Runtime)Classpath$ -          
           Dorg.gradle.parallel= -Dorg.gradle.console=plain -PonlySubProject=app -I     
          /tmp-585-FzuNOJ9t3n3t--init.gradle --no-configuration-cache  


 FAILURE: Build failed with an exception.  
                                   
* What went wrong:                                                           
    Configuration Cache cannot be disabled when Isolated Projects is enabled. 
```

It is expected that this will make it difficult to use the Snyk CLI in large Gradle-based projects.

PR #351, opened 2026-08-31
Better error messaging until Gradle IP support is implemented

trevjonez:
### What this does

On Gradle 7+ this plugin passes `--no-configuration-cache`. Isolated Projects is built on the configuration cache and cannot run without one, so Gradle rejects the combination outright and the scan dies before any dependency resolution happens:

```
* What went wrong:
Configuration Cache cannot be disabled when Isolated Projects is enabled.
```

Nothing in that message mentions Snyk, and the flag it names is one the user never passed. It reads as a Gradle problem with their own build, so the natural next step is hunting through `gradle.properties` rather than looking for a Snyk limitation. This adds a message that names the cause, links the tracking issue, and gives a working per-invocation override.

Related issue: https://github.com/snyk/snyk-gradle-plugin/issues/344

Clarity only — this does not add Isolated Projects support.

Verified on Gradle 9.5.1 against `test/fixtures/multi-project`, adding one line to that fixture's `gradle.properties`:

```
org.gradle.unsafe.isolated-projects=true
```

Reproduce the failure the marker matches (this is the plugin's own invocation shape):

```
gradle snykResolvedDepsJson --project-dir test/fixtures/multi-project \
  -I "$(pwd)/lib/init.gradle" -PonlySubProject=. \
  --no-configuration-cache --no-daemon
```

Then confirm the override the message suggests does let a scan complete, with the property still `true`:

```
gradle snykResolvedDepsJson --project-dir test/fixtures/multi-project \
  -I "$(pwd)/lib/init.gradle" -PonlySubProject=. \
  -Dorg.gradle.unsafe.isolated-projects=false \
  --no-configuration-cache --no-daemon
```

That emits `JSONDEPS {...}` and `BUILD SUCCESSFUL`, so the advice is tested rather than assumed. Revert the fixture edit afterwards.

---

> [!NOTE]
> **Low Risk**
> User-facing error messaging and a guarded regex branch only; no change to Gradle invocation or dependency resolution behavior.
> 
> **Overview**
> When Gradle fails because **Isolated Projects** conflicts with the plugin’s **`--no-configuration-cache`** flag (Gradle 7+), users now get a dedicated error instead of the generic “check your arguments” message.
> 
> The change adds a **case-insensitive regex** on the stable core of Gradle’s wording (covers both pre- and post-9.5 phrasing), wires it as an **`else if`** after the Android variant-resolution branch so precedence stays explicit, and surfaces guidance: unsupported limitation, link to issue **#344**, and per-scan overrides via **`snyk test -- -Dorg.gradle.unsafe.isolated-projects=false`** (or the non-`unsafe` property on newer Gradle). The pattern is exported for tests; functional specs assert match against real Gradle messages and no false positives on other configuration-cache errors.
> 
> **No Isolated Projects support** is added—only clearer diagnostics and workaround instructions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 45cc3765a04385ff6f56f200d1b6b9577542357c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -40,6 +40,14 @@ const cannotResolveVariantMarkers = [
   'Unable to find a matching variant of project',
 ];
 
+// Gradle's own wording when --no-configuration-cache (which this plugin passes on Gradle 7+)
+// meets a build with Isolated Projects enabled. Isolated Projects is built on the configuration
+// cache and cannot run without one, so the two are mutually exclusive and the build fails before
+// any dependency resolution happens.
+const isolatedProjectsMarkers = [
+  'Configuration Cache cannot be disabled when Isolated Projects is enabled',
+];
+
 type Options = api.InspectOptions & GradleInspectOptions & CliOptions;
 type VersionBuildInfo = api.VersionBuildInfo;
 
@@ -559,6 +567,35 @@ to
     ${chalk.whiteBright(
       "implementation project(path: ':mymodule', configuration: 'default')",
     )}`;
+    }
+
+    // Checked after the variant case: when Isolated Projects is what failed, nothing else in the
+    // build got far enough to be the real cause, so this message should win.
+    if (isolatedProjectsMarkers.find((m) => error.message.includes(m))) {
+      mainErrorMessage = `Error running Gradle dependency analysis.
+
+Your build has Gradle Isolated Projects enabled, which this plugin does not support yet.
+
+Isolated Projects is built on the configuration cache and cannot run without one, while this
+plugin passes ${chalk.whiteBright(
+        '--no-configuration-cache',
+      )} on Gradle 7 and above. Gradle rejects that
+combination before any dependency resolution happens, so the scan cannot start.
+
+Progress is tracked in ${chalk.whiteBright(
+        'https://github.com/snyk/snyk-gradle-plugin/issues/344',
+      )}.
+
+To scan in the meantime, turn Isolated Projects off for the scan. It is a Gradle property, so it
+can be overridden per invocation without editing your build:
+    ${chalk.whiteBright(
+      'snyk test -- -Dorg.gradle.unsafe.isolated-projects=false',
+    )}
+
+Use ${chalk.whiteBright(
+        '-Dorg.gradle.isolated-projects=false',
+      )} instead if your Gradle version has dropped the
+"unsafe" prefix from the property name.`;
     }
 
     error.message = `${chalk.red.bold(


lib/index.ts
```diff
@@ -40,6 +40,14 @@ const cannotResolveVariantMarkers = [
   'Unable to find a matching variant of project',
 ];
 
+// Gradle's own wording when --no-configuration-cache (which this plugin passes on Gradle 7+)
+// meets a build with Isolated Projects enabled. Isolated Projects is built on the configuration
+// cache and cannot run without one, so the two are mutually exclusive and the build fails before
+// any dependency resolution happens.
+const isolatedProjectsMarkers = [
+  'Configuration Cache cannot be disabled when Isolated Projects is enabled',
```
gergo-papp: This detection is an exact substring match against one specific Gradle wording. If a supported Gradle version phrases the failure differently (different casing, e.g. "configuration cache" vs "Configuration Cache", or reworded text — Gradle has changed this kind of message wording across major versions before), `isolatedProjectsMarkers.find(...)` returns undefined and the new friendly message never fires. Users then silently fall back to the generic "Please ensure you are calling the `snyk` command with correct arguments" message for exactly the scenario this PR was written to explain, with nothing surfacing the mismatch. Worth confirming this string is stable across the Gradle versions this plugin supports (or matching on a more version-resilient fragment).

---
_Generated by [Claude Code](https://claude.ai/code)_

lib/index.ts
```diff
@@ -561,6 +569,35 @@ to
     )}`;
     }
 
+    // Checked after the variant case: when Isolated Projects is what failed, nothing else in the
+    // build got far enough to be the real cause, so this message should win.
+    if (isolatedProjectsMarkers.find((m) => error.message.includes(m))) {
```
gergo-papp: Precedence between this Isolated-Projects case and the Android-variant case above is enforced only by source order plus the comment on the previous two lines — both blocks unconditionally overwrite the same `mainErrorMessage` variable rather than using `else if`/early-return. That's fine today since the marker sets can't realistically both match, but it means a future contributor who reorders these blocks, or inserts a third marker-check between them, can silently flip which message wins with no test to catch the regression. Consider driving this off an ordered array of `{markers, buildMessage}` and picking the first match, so precedence is structural rather than comment-enforced.

---
_Generated by [Claude Code](https://claude.ai/code)_

lib/index.ts
```diff
@@ -40,6 +40,14 @@ const cannotResolveVariantMarkers = [
   'Unable to find a matching variant of project',
 ];
 
+// Gradle's own wording when --no-configuration-cache (which this plugin passes on Gradle 7+)
+// meets a build with Isolated Projects enabled. Isolated Projects is built on the configuration
+// cache and cannot run without one, so the two are mutually exclusive and the build fails before
+// any dependency resolution happens.
+const isolatedProjectsMarkers = [
```
gergo-papp: No automated test was added for this new branch (mirrors the pre-existing convention of not testing the Android-variant case either, but that one is at least flagged in-code as "no automated tests for this yet"). The PR description documents a manual repro/verification against a live Gradle install, which is great, but nothing in CI pins the marker string or the resulting message, so a later refactor of the `getAllDeps` catch block (e.g. changing marker text or reordering the two special cases) could regress this silently. Consider a small unit test that feeds a synthetic error containing the marker text through this code path and asserts on `mainErrorMessage`/`error.message`.

---
_Generated by [Claude Code](https://claude.ai/code)_

gergo-papp (commented):
Automated review (high effort, recall-biased). This is a small, additive, clearly-scoped change (better error messaging for the Gradle Isolated Projects conflict). No hard correctness bugs found; left three comments on robustness/maintainability of the new special-case branch — see inline.

---
_Generated by [Claude Code](https://claude.ai/code)_

lib/index.ts
trevjonez: Confirmed, and worse than described: verified directly against real Gradle output that the wording actually differs within the 8.8+ range where Isolated Projects exists at all.

    8.13 / 8.14.3 / 9.0.0: "The configuration cache cannot be disabled when isolated projects is enabled."
    9.5.1 / 9.7.1:         "Configuration Cache cannot be disabled when Isolated Projects is enabled"

The old exact-case literal only matched the second form, so on 8.8-9.4 a real build hitting this fell into the generic fallback message instead. Replaced with a case-insensitive regex on the invariant core of both sentences, and reproduced end-to-end against a real Gradle 8.14.3 build to confirm the friendly message now fires there. Pushed in 45cc376.

lib/index.ts
trevjonez: Agreed. Changed to `else if` rather than a second independent `if`, so precedence against the variant case above is a language guarantee, not "whichever block runs last wins the overwrite." Didn't go as far as the ordered-array/first-match-wins table you sketched - for two cases `else if` is the same guarantee with less machinery; happy to revisit if a third case gets added and the chain starts feeling unwieldy. Pushed in 45cc376.

lib/index.ts
trevjonez: Added functional tests for the marker pattern (exported as `isolatedProjectsErrorPattern`) against both real wordings found above, plus a negative case so an unrelated configuration-cache message doesn't false-match. Stopped short of a full system test that runs a real Gradle build through `inspect()`: existing tests in this style (`multi-module.test.ts`) gate on a runtime-detected Gradle major version using a plain `if` inside the `describe` body, evaluated before the async `beforeAll` that sets that variable has run - so as far as I can tell those gated tests never actually register. Building a correctly-gated version felt like scope beyond this PR, especially since #352 makes this whole error path unreachable on any Gradle version that can enable Isolated Projects at all. Pushed in 45cc376.

PR #352, opened 2026-09-01
feat: support the configuration cache and Isolated Projects on Gradle 8.1+

### LINKED ISSUE #344: Support configuration-cache/isolated-project feature

trevjonez:
### Human part of the PR body

Decided I would give Opus a crack at this feature since it was going to be a big lift and this project has pretty good
test coverage making it a prime candidate for automated work. The new init scripts should leave this project in great
shape for when Gradle 10.0 ships and improve build times for those using more modern versions even though that wasn't
the goal.

I don't love the split script it came up with but unless you cut off support for most of the version matrix there
doesn't seem to be a good way to do it.

### AI part of the PR body (human edited)

- [x] Documentation written
- [x] Commit history is tidy

### What this does

Adds configuration-cache and Isolated Projects support for Gradle 8.1+, and stops forcing
`--no-configuration-cache` there.

Closes #344. And would make #351 moot.

To be explicit about scope: **this does not turn the configuration cache on.** It stops turning it off. Whatever the
build is already configured to do is what runs, and a caller-supplied
`--configuration-cache` simply stops being stripped. `buildArgs` never pushes `--configuration-cache`.

Why forcing it off is not neutral: Isolated Projects is built on the configuration cache and cannot run without one, so
Gradle rejects the combination outright and the scan dies before resolving anything

```
* What went wrong:
Configuration Cache cannot be disabled when Isolated Projects is enabled.
```

Nothing in that message mentions Snyk, and the flag it names is one the user never passed.

Below 8.1 nothing changes: `lib/init.gradle` is byte-for-byte untouched and still gets `--no-configuration-cache`.

### Why a second script rather than editing `init.gradle`

Groovy resolves types at **compile** time, so naming a class absent from the running Gradle fails the whole script at
parse time, before any version gate can execute:

```
initialization script: unable to resolve class com.example.totally.Absent
     class ReferencesMissingInterface implements com.example.totally.Absent {
  2 errors
```

Class linking is not lazy. `BuildService` (Gradle 6.1) and the lazy `ResolutionResult.getRootComponent()` (Gradle 7.4)
therefore cannot appear in a file that must also parse on Gradle 4.10, which CI still covers. One file cannot do both
jobs.

`IsolatedAction` (8.8) is never named even in the new file, for the same reason, see the bootstrap.

| Gradle  | script               | configuration cache      | Isolated Projects | parallel execution       |
|---------|----------------------|--------------------------|-------------------|--------------------------|
| 4 – 8.0 | `lib/init.gradle`    | forced off               | not supported     | forced off               |
| 8.1+    | `lib/init-cc.gradle` | left as the build has it | supported on 8.8+ | left as the build has it |

### Why the floor is 8.1 and not 7.4

The APIs land in 7.4, but 7.4.x cannot store a cache entry for a build that has an unresolvable dependency. The resolved
graph carries the failure exception, and serializing it dies on JDK 17+:

```
field 'causes' from type 'ModuleVersionNotFoundException': error writing value of type
'java.util.concurrent.CopyOnWriteArrayList'
> Unable to make private void CopyOnWriteArrayList.writeObject(...) accessible:
  module java.base does not "opens java.util.concurrent" to unnamed module
```

Gradle's own `dependencies` task fails identically on `test/fixtures-with-wrappers/basic-with-failed-dep` under 7.4.2,
and succeeds on 8.1.1 and every version after, so this is upstream rather than something to work around. Setting the
floor above it avoids shipping a version that breaks scans which currently succeed. 8.1 is also where the configuration
cache became stable.

CI already covers both sides of the boundary: `7.3` exercises the legacy file and `8.4` the new one. Nothing exercises
7.4–8.0 on the new path, which is fine because they stay on the legacy one.

### How it works

The old script has one task walk `rootProject.allprojects` at execution time, reading every project's configurations
from whichever project's task happened to run first. That is both a configuration-cache violation (`Task.project` at
execution time) and an Isolated Projects violation (cross-project model access).

Now each project reports only itself into a shared `BuildService`, and the service's `close()` prints the merged result
once at the end of the build. A build service fits because it needs **no task ordering** so there are no cross-project
task references, which is what keeps it IP-clean.

**This also lets `-Dorg.gradle.parallel=` go on 8.1+.** That flag was there because
`init.gradle` emits from a task action walking every project, which a parallel build can race into producing more than
one result line. That is now structurally impossible, so builds configured for parallel execution get it.

Worth noting for anyone reasoning about this: Isolated Projects does **not** implicitly enable parallel project
execution. Measured via `startParameter.isParallelProjectExecutionEnabled()` on Gradle 9.5.1, with a fresh cache entry
per run:

| Isolated Projects | flag                     | `parallelProjectExecution` |
|-------------------|--------------------------|----------------------------|
| off               | none                     | false                      |
| off               | `-Dorg.gradle.parallel=` | false                      |
| off               | `--parallel`             | true                       |
| on                | none                     | false                      |
| on                | `-Dorg.gradle.parallel=` | false                      |
| on                | `--parallel`             | true                       |

What IP parallelizes is project *configuration*, a different axis which is the other reason the flag was not pulling its
weight there, since configuration is where the injected script does its wiring and IP runs it concurrently regardless of
this setting.

That also settles the existing TODO at `lib/index.ts:180`:

> the longer term fix here is to make multiple snyk graph outputs impossible by interacting with
> Gradle API in a more consistent way

Exactly one thing prints, so multiple `JSONDEPS` lines are now impossible by construction rather than guarded against
after the fact. No parsing changes were needed.

Build-level facts a single project cannot see (`defaultProject`, `allSubProjectNames`) are derived from the settings
model in `settingsEvaluated` and passed in as plain strings. The invoked project is now matched from `startParameter`,
rather than inferred from whichever task ran first.

### The part most worth reviewing

Task state holds **plain data only** never a resolution handle. The configuration cache realizes a provider reachable
from task state while storing the entry, so storing `Provider<ResolvedComponentResult>` means storing Gradle's resolved
graph, including `UnresolvedDependencyResult.failure` for anything that failed to resolve. Those are arbitrary exception
types:

```
field `cause` of `SSLHandshakeException` ... field `rootComponent` of `SnykConfigurationHandle`
... error writing value of type 'sun.security.validator.ValidatorException'
> Unable to make field ... accessible: module java.base does not "opens sun.security.validator"
```

So the whole per-project walk happens inside one provider that yields a plain `Map`, and the task holds only a
`MapProperty`. One provider rather than one per configuration, because every configuration contributes to a single
shared graph and node-id map folding N providers would leave that accumulation order up to Gradle.

Node ids are scoped per project rather than per build, which is sound because `lib/graph.ts` turns each project's
`gradleGraph` into its own dep-graph. The `sha1Map` is still merged build-wide; a normalized id is the SHA-1 of the
artifact's own bytes, so two projects sharing a dependency agree and the merge cannot conflict.

### ⚠️ Works around an upstream Isolated Projects bug, deliberately

`SnykGraphBuilder.projectGroup` trims a trailing dot from a **project** component's group.

Gradle 8.11-ish through 8.14.x append one to a subproject's default `project.group` when Isolated Projects is enabled,
so
`:list` arrives as `root.:list` rather than `root:list`. That is
[gradle/gradle#33248](https://github.com/gradle/gradle/issues/33248), fixed by
[#34268](https://github.com/gradle/gradle/pull/34268) in **Gradle 9.1.0** and not backported, so the newest 8.x line
still carries it. The cause is the identity check `getParent() == rootProject` in `DefaultProject.getGroup()`, which
does not hold under isolation; the direct child then takes the else branch and appends
`"." + parent.getPath().substring(1)`, empty for the root. Measured with `project.group` printed from a `beforeProject`
hook, root named `demo-root`:

| Gradle         | `:list` group with IP on |
|----------------|--------------------------|
| 8.10.1         | `demo-root`              |
| 8.13           | `demo-root.`             |
| 8.14.3         | `demo-root.`             |
| 9.2.1 to 9.7.1 | `demo-root`              |

Only direct children are affected; a nested `:a:b` reports `demo-root.a` either way.

Untrimmed, every project-dependency node id shifts the moment a build turns Isolated Projects on. Nothing errors out,
since `parseCoordinate` splits on `:` and `@` and a trailing dot is inert, but downstream it reads as all of that
project's dependencies being replaced, and anything keyed to the previous id stops matching. Trimming keeps a scan
reporting the same coordinates whichever way the build is configured.

The trim is narrow on purpose: project components only, and only a trailing dot. It cannot rewrite an external
coordinate, and it is a no-op on versions without the bug.

Worth being precise about one edge, since the bug only affects a *default* group and nothing at that point can tell a
defaulted group from one the build set explicitly: a project declaring `group = 'com.example.'` gets trimmed too. That
is reachable only by declaring a group whose trailing dot already makes it an invalid coordinate, which seems better to
normalize than to preserve, but it is a deliberate choice rather than an accident.

The Isolated Projects test asserts against the same golden file as the cache-only tests with nothing normalized, on a
fixture pinned to 8.13, so the trim is what makes it pass.

### Verification

**Output parity against `init.gradle`, semantically diffed** (key order normalized, and `parentIds` sorted since it is a
`Set` on the Gradle side). Every fixture in `test/fixtures` and `test/fixtures-with-wrappers`, each in
`-PonlySubProject=.` and all-projects form, and the new script run with the cache both **on and off**:

```
IDENTICAL: 232    DIFFERS: 0
```

Fixtures pinned below 8.1 are reported as skipped, since they use the legacy script. The handful of no-output cases are
all the *legacy* script failing on this machine for pre-existing reasons (`malformed-build-gradle` is meant to fail;
`gradle-kts` and `java-reachability-playground` don't configure on 9.5).

**Cache / Isolated Projects matrix** on `test/fixtures/multi-project`, with
`-Dorg.gradle.configuration-cache.problems=fail`:

| Isolated Projects | configuration cache | result                                                                                        |
|-------------------|---------------------|-----------------------------------------------------------------------------------------------|
| off               | on                  | 1 JSONDEPS                                                                                    |
| off               | off                 | 1 JSONDEPS                                                                                    |
| off               | unset               | 1 JSONDEPS                                                                                    |
| on                | on                  | 1 JSONDEPS, zero problems                                                                     |
| on                | unset               | 1 JSONDEPS                                                                                    |
| on                | explicitly off      | Gradle refuses the build asked for a contradiction, and the plugin no longer passes that flag |

Also verified: `close()` still prints on a configuration-cache **hit**, which is the case where an init-script-level
hook would not run at all.

**Automated coverage.** `test/system/configuration-cache.test.ts` runs a real Gradle build per case. Every one of them
passes `-Dorg.gradle.configuration-cache.problems=fail`, so a regression fails a test instead of quietly degrading to no
caching:

| test                                | what it pins down                                                                                         |
|-------------------------------------|-----------------------------------------------------------------------------------------------------------|
| cache enabled                       | asserts `.gradle/configuration-cache` exists, so a silent fallback fails                                  |
| second scan reuses the stored entry | a cache hit does not re-run the init script, so the tasks and collector service are restored, not rebuilt |
| Isolated Projects enabled           | the 8.8+ `gradle.lifecycle.beforeProject` bootstrap                                                       |
| cache enabled on an 8.4 fixture     | the 8.1 to 8.7 `gradle.allprojects` bootstrap, which nothing else exercises                               |
| cache explicitly disabled           | the script must not depend on the cache being on, which is what most builds still do                      |

The first three plus the last use the `init-gradle` fixture, whose wrapper pins 8.13, so they sit above both thresholds
whatever version the CI matrix is running. The 8.4 case compares its own cache-on and cache-off runs rather than a
golden file, which asserts the invariant that matters and adds no fixture to maintain.

**Suites:** `npx tsc` clean, `npm run lint` clean (0 errors; the 19 warnings are pre-existing and untouched), Prettier
clean, `test:functional` 8/8 suites and 88 tests (18 new, covering script selection and both flags at both tiers),
`test:system` 16/16 suites and 94 tests (5 new).

### Gradle 10 readiness, and one other fix

Two things already in the code would have broken on the next major:

- `getAllDeps` rejected Gradle 1.x with `gradleVersion.match(/Gradle 1/)`, which also matches `Gradle 10` and every
  later two-digit major. Now anchored as `/Gradle 1\./`.
- `DomainObjectCollection.findAll(Closure)` is deprecated in 9 and scheduled for removal in 10. Gradle's own problems
  report flagged it against the new script, so it snapshots `project.configurations` to a list before filtering, which
  hands the work to Groovy's `List.findAll` instead.

Unrelated but adjacent: `.gitignore` covered `test/fixtures/**/build` but not `test/fixtures-with-wrappers/**/build`, so
running `test:system` left untracked build output behind.

### Known limitations, stated plainly

- 7.4–8.0 keep the old behaviour, for the resolution-failure reason above.
- With `-PsnykIncludeComponentMetadata`, `distribution:url` needs a build-operation listener that can only be registered
  while the init script runs, and an init script does not run on a cache hit. Such a build carries no
  `distribution:url`. Close to moot in practice a cache hit against a warm dependency cache performs no resource reads,
  so there would be nothing to capture and artifact hashes are unaffected because they are driven by a task input. The
  old `gradle.buildFinished` cleanup, itself a cache problem, is replaced by the collector service's `close()`.

### Notes for the reviewer

**Duplication is deliberate.** The new file repeats the graph walker rather than sharing it, because sharing is exactly
what the Groovy compile-time constraint above forbids. It is self-contained, so `init.gradle` can simply be deleted when
support for < 8.1 is dropped, leaving one implementation.

For whenever that gets weighed: per `gh api repos/gradle/gradle/security-advisories`, only the `8.14.x` line still
receives security backports. `4.10.3` (Dec 2018), `5.6.4` (Nov 2019), `6.9.4` (Feb 2023), `7.6.6` (Jul 2025) and `8.0.2`
(Mar 2023) are all end-of-life under Gradle's own policy, and every one of them is inside the affected range of
CVE-2026-22865 and CVE-2026-22816 (both high) with no in-line fix. Supporting Gradle 4.x is what costs this repo the
second file. Might be a good idea in the future for this plugin to also report that back to the console as an issue to
fix as well as warn in the CLI when the dep dump is happening.

**A pre-existing comment was wrong and I dropped rather than reproduced it.** `buildArgs` said the configuration cache
"is enabled by default in Gradle 9". It is not the 9.0 release notes call it "the preferred mode" and "not yet
required", and describe silencing the suggestion with
`org.gradle.configuration-cache=false`, which only makes sense if the default is off.

```bash
# Isolated Projects, which fails on main before this change
cd test/fixtures/multi-project
gradle snykResolvedDepsJson -q -I ../../../lib/init-cc.gradle -PonlySubProject=. \
  --no-daemon --configuration-cache \
  -Dorg.gradle.unsafe.isolated-projects=true \
  -Dorg.gradle.configuration-cache.problems=fail
```

Expect one `JSONDEPS` line and `BUILD SUCCESSFUL`. Swap `lib/init-cc.gradle` for `lib/init.gradle` to see the failure
this replaces. Requires Gradle 8.8+ for the Isolated Projects part; 8.1+ for the cache alone.

---

> [!NOTE]
> **Medium Risk**
> Core Gradle scan behavior changes for 8.1.1+ (cache, parallel, new ~1.2k-line init script), though legacy Gradle is untouched and coverage is extensive; mis-version detection would pick the wrong script.
> 
> **Overview**
> Adds **configuration-cache and Isolated Projects support** for Gradle **8.1.1+** by introducing a second injected script, `init-cc.gradle`, while **`init.gradle` stays unchanged** for older Gradle.
> 
> **Node layer (`lib/index.ts`)** picks the script from `gradle -v` (patch-aware so **8.1.0** stays on the legacy path). On 8.1.1+ it **stops forcing** `--no-configuration-cache` and `-Dorg.gradle.parallel=`, which fixes scans when Isolated Projects requires the cache and removes the old parallel workaround now that output is merged via a single build service. Below the floor, behavior is unchanged. Also fixes Gradle **10+** being rejected by the old `/Gradle 1/` check.
> 
> **`init-cc.gradle`** restructures collection: each project reports into a shared **`BuildService`** that prints one `JSONDEPS` line on `close()`, uses lazy resolution mapped to **plain data** for cache safety, and bootstraps per-project wiring via `settingsEvaluated` (`beforeProject` on 8.8+, `allprojects` on 8.1.1–8.7). Includes a narrow **trailing-dot trim** on project dependency groups for an Isolated Projects Gradle bug.
> 
> Docs, packaging, `.gitignore`, expanded **functional** tests for flags/script selection, and new **system** tests for cache reuse, Isolated Projects, and both bootstrap tiers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6b86619777be4f71c7f48114b5f5165ab0e14da6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/index.ts
```diff
@@ -390,6 +401,59 @@ function getVersionBuildInfo(
...
+// needs land earlier (ResolutionResult.getRootComponent() and
+// ArtifactCollection.getResolvedArtifacts() are both 7.4), but 7.4.x cannot
+// actually store an entry for a build that has an unresolvable dependency: the
+// resolved graph carries the failure exception, and serializing it dies on JDK 17+
+// ("module java.base does not open java.util.concurrent"). Gradle's own
+// `dependencies` task fails identically there, and 8.1.1 onwards does not, so the
+// floor sits above it rather than shipping a version that breaks scans which
+// currently succeed.
+const CC_INIT_SCRIPT_MIN_MAJOR = 8;
+const CC_INIT_SCRIPT_MIN_MINOR = 1;
```
gergo-papp, line 419: **Correctness: the floor accepts bare `Gradle 8.1` (i.e. 8.1.0), which the PR's own comment says still crashes.**

The comment right above this (lines 410-417) explains the floor exists because 7.4.x–8.1.0 dies serializing an unresolvable-dependency cache entry on JDK 17+, and that "8.1.1 onwards does not" fail. But `parseGradleVersion` (line 427) only captures major.minor, dropping any patch component, and `supportsConfigurationCache` (line 444-448) accepts `minor >= CC_INIT_SCRIPT_MIN_MINOR` (`1`). So a project reporting exactly `Gradle 8.1` (the initial 8.1.0 GA release) parses to `{major: 8, minor: 1}` and is treated as cache-compatible — the exact version the comment says still hits the JDK17 serialization crash.

Concretely: a build on Gradle 8.1.0 with any unresolvable dependency, on JDK 17+, gets routed to `init-cc.gradle` with the configuration cache left on, instead of falling back to `init.gradle` + `--no-configuration-cache` as before. Since `gradle -v` never reports a patch version in its "Gradle X.Y" banner line, there's no patch-aware signal to gate on from this string alone — the fix is probably to require `minor > 1` (i.e. floor at 8.2) rather than `>= 1`, unless there's a way to special-case 8.1.x as unsupported entirely.

---
_Generated by [Claude Code](https://claude.ai/code)_

lib/init-cc.gradle
```diff
@@ -0,0 +1,1176 @@
...
+// URLs via an internal build-operation listener. On a warm cache nothing is
+// re-fetched, so a node simply carries no distribution:url: graceful omission,
+// never a reconstructed guess.
+
+class SnykResourceUrlIndex {
+
+    // filename -> Set<url>. A Set because a shared cache can see the same
+    // filename from more than one repo across a build; disambiguated at lookup
+    // by Maven layout.
+    private static final Map<String, Set<String>> URLS_BY_FILENAME = new ConcurrentHashMap<>()
```
gergo-papp: **`URLS_BY_FILENAME` is a JVM `static` field and is never cleared, unlike the listener it shares a lifecycle comment with — it leaks across builds on a reused daemon and can hand back stale/wrong URLs.**

The old `lib/init.gradle` held the equivalent map as a script `@Field` (line 180 there), which is a fresh instance every build. Turning it into `static final` here (line 142) ties its lifetime to the loaded class, i.e. the daemon, not one build — the exact concern this class's own comment raises for `listenerManager`/`listener` two lines down ("a listener left registered survives into later builds on a reused daemon, accumulating"). `unregister()` (lines 182-194) nulls `listenerManager`/`listener` but never calls `URLS_BY_FILENAME.clear()`.

Two concrete costs on a reused `--daemon` process:
1. Unbounded memory growth — one entry accumulates per distinct artifact filename across every scan run on that daemon, forever.
2. Correctness: this file explicitly documents `distribution:url` as "TRUE provenance only... graceful omission, never a reconstructed guess" (lines 131-135). But if scan N's cache is warm and no resource read happens for a filename that scan N-1 *did* read (and record), `lookup()` (line 282-289) still returns scan N-1's stale URL, since nothing scopes the map to the current build. That silently violates the file's own provenance guarantee.

Suggest clearing `URLS_BY_FILENAME` inside `unregister()`'s `finally` block alongside the listener reset.

---
_Generated by [Claude Code](https://claude.ai/code)_

lib/init-cc.gradle
```diff
@@ -0,0 +1,1176 @@
...
+        return matches
+    }
+}
+
+// ─── bootstrap ──────────────────────────────────────────────────────────────
+
+// Component metadata capture has to be armed before resolution begins, so it is
+// set up here rather than from a task. Gate on the value, not mere presence, so
+// an explicit =false disables the feature.
+if (gradle.startParameter.projectProperties.get('snykIncludeComponentMetadata')?.toBoolean()) {
```
gergo-papp: **`snykIncludeComponentMetadata` is gated here on a narrower property source than the one that actually drives the feature per-project, so `gradle.properties`-only configuration silently gets a partial result.**

This bootstrap check reads `gradle.startParameter.projectProperties`, which per Gradle only reflects `-P` command-line arguments. But `readScanSpec` (line 951) reads the same-named property via `property(project, 'snykIncludeComponentMetadata')`, which resolves through `project.providers.gradleProperty(...)` — and the comment right above `readScanSpec` (lines 930-937) explicitly says this is deliberate, *specifically* "so that values set in `gradle.properties` are honoured too."

Concretely: a team that sets `snykIncludeComponentMetadata=true` only in a committed `gradle.properties` (no `-P` on the CLI) will have `readScanSpec` compute `includeComponentMetadata: true` and emit artifact hashes accordingly, but `SnykResourceUrlIndex.register(gradle)` never gets called at line 1104 (since `projectProperties` doesn't see the `gradle.properties` value), so the resource-read listener never arms and `distribution:url` is silently always absent — indistinguishable from the "graceful omission on a cache hit" case this file otherwise documents as expected.

Since the surrounding code already went out of its way to make `gradleProperty` the source of truth for this exact CLI-vs-file distinction, this bootstrap check should probably use the same source (e.g. read it once via a `providers.gradleProperty` lookup against `gradle` /the root project, rather than `startParameter.projectProperties`) so the two gates can't disagree.

---
_Generated by [Claude Code](https://claude.ai/code)_

lib/init-cc.gradle
```diff
@@ -0,0 +1,1176 @@
...
+    // filename from more than one repo across a build; disambiguated at lookup
+    // by Maven layout.
+    private static final Map<String, Set<String>> URLS_BY_FILENAME = new ConcurrentHashMap<>()
+
+    // Held so the listener can be removed again. BuildOperationListenerManager
+    // is a daemon-scoped service, so a listener left registered survives into
+    // later builds on a reused daemon, accumulating one per build and pinning
+    // each build's captured state.
+    private static Object listenerManager = null
+    private static Object listener = null
```
gergo-papp: **`listenerManager`/`listener` are single static slots, but `BuildOperationListenerManager` is daemon-scoped and this init script registers once per Gradle instance in the invocation — a `buildSrc` or included build can overwrite the root build's reference before it's unregistered.**

`register(gradle)` (line 157) is invoked once per Gradle instance this `-I` script is applied to, and per the comment on line 144-147 that manager is shared across the whole daemon (main build, `buildSrc`, included builds). Each `register()` call unconditionally overwrites the static `listenerManager`/`listener` pair rather than tracking one per Gradle instance. If a `buildSrc` or `includeBuild` build's Gradle instance registers and finishes (calling `unregister()`, which nulls the statics) *after* the root build has already registered but *before* the root build's own `SnykDepsCollector.close()`/`unregister()` runs, the root build's `unregister()` call later finds `listenerManager`/`listener` already null and never actually calls `removeListener` for its own, still-live listener. That listener then stays registered on the daemon-scoped manager for the rest of the daemon's life — the "bounded per-build leak" the comment at line 188-189 assumes doesn't hold in that ordering, and it compounds every scan run against a repo with `buildSrc`/composite builds on a reused daemon.

Worth tracking multiple registrations (e.g. a list/map keyed by the `gradle` instance) rather than a single overwritable slot, if `buildSrc`/composite-build scans with `-PsnykIncludeComponentMetadata` are expected to work correctly on a long-lived daemon.

---
_Generated by [Claude Code](https://claude.ai/code)_

lib/index.ts
```diff
@@ -697,10 +764,22 @@ function buildArgs(
...
+  // matter how many projects report concurrently.
+  //
+  // Under Isolated Projects the flag was never protecting the phase that matters
+  // anyway: project configuration is parallel there regardless of this setting,
+  // and configuration is where the injected script does its wiring.
+  //
+  // Gradle 4.3.0+ has a `--no-parallel` flag, but older versions need this form.
+  // Not `=false`, to stay compatible with 3.5.x:
+  // https://github.com/gradle/gradle/issues/1827
+  if (!supportsConfigurationCache(gradleVersion)) {
```
gergo-papp: Minor: `supportsConfigurationCache(gradleVersion)` is called three times in this function (here, and again at lines 812 and 835) for the same input, each re-running `parseGradleVersion`'s regex. Not a real perf issue, but it means three call sites have to keep agreeing on the same boolean — worth hoisting into a local (`const supportsCC = supportsConfigurationCache(gradleVersion)`) near the top of `buildArgs` and reusing it, so a future change to the gating logic can't desync the parallel/no-cc/unsupported-args decisions.

---
_Generated by [Claude Code](https://claude.ai/code)_

lib/init-cc.gradle
```diff
@@ -0,0 +1,1176 @@
...
+    private static void wire(Project project, SnykCollectDepsTask task, Map spec) {
+        task.inScope.set((boolean) spec.inScope)
+
+        // toList() before findAll, deliberately. DomainObjectCollection.findAll(Closure)
+        // is deprecated in Gradle 9 and scheduled for removal in Gradle 10, so calling it
+        // straight on project.configurations would break this script on the next major.
+        // Snapshotting first hands the filtering to Groovy's own List.findAll, and a
+        // snapshot is what this wants regardless, since the result is iterated twice.
+        List<Configuration> matching = project.configurations.toList().findAll { conf ->
+            conf.name =~ spec.confNameFilter && matchesAttributeFilter(conf, spec.confAttrSpec)
```
gergo-papp: Minor: when `-PconfAttr` is set, `matchesAttributeFilter` (called here, defined at line 1079) already iterates every attribute of every matching configuration and calls `getAttribute(attr).toString()` on each. Then `matching.each` at line 1017-1027 walks the same configurations' `attributes.keySet()` again and calls `getAttribute(attr).toString()` again to populate `observedAttributes`. That's every matching configuration's attributes stringified twice (including re-realizing any lazy/provider-backed attribute value — the operation the OSM-3867 comment flags as CME-prone). Could have `matchesAttributeFilter` return the name→stringValue snapshot it already builds and feed both the filter decision and `observedAttributes` from one pass.

---
_Generated by [Claude Code](https://claude.ai/code)_

gergo-papp (commented):
High-effort review of the configuration-cache/Isolated Projects support (6 parallel finder agents covering `lib/index.ts`, `lib/init-cc.gradle`, removed-behavior vs. `init.gradle`, cross-file call-site tracing, cleanup/efficiency, and root-cause assessment), each finding verified against the actual diff before posting.

Overall this is a well-reasoned, well-documented PR — the two-script split, the `BuildService`-based single-print design, and the disclosed known limitations (no `distribution:url` on a cache hit, the `projectGroup` trailing-dot workaround, deliberate duplication between the two scripts) all check out and are already explained in the PR description. The comments below are the gaps that aren't already covered by that description:

**Correctness (verified against the diff):**
1. `lib/index.ts` — the configuration-cache floor (`CC_INIT_SCRIPT_MIN_MINOR = 1`) accepts bare `Gradle 8.1`, but the PR's own stated rationale is that only 8.1.1+ avoids the JDK17 serialization crash; `parseGradleVersion` drops the patch component entirely, so 8.1.0 gets routed onto the new cache-compatible path anyway.
2. `lib/init-cc.gradle` — `SnykResourceUrlIndex.URLS_BY_FILENAME` is a `static final` map (unlike the old script's per-build `@Field`) and is never cleared in `unregister()`, unlike the listener it sits next to. Leaks across builds on a reused daemon and can return a stale/wrong URL, contradicting the file's own "TRUE provenance only" comment.
3. `lib/init-cc.gradle` — the bootstrap gate for `snykIncludeComponentMetadata` reads `gradle.startParameter.projectProperties` (CLI-only), while `readScanSpec` deliberately reads the same property via `gradleProperty` (CLI + `gradle.properties`) — a config set only in `gradle.properties` silently gets hashes but no `distribution:url`, with no error.
4. `lib/init-cc.gradle` — `listenerManager`/`listener` are single static slots shared across every Gradle instance the init script is applied to in one invocation (root + `buildSrc` + included builds); an unlucky ordering lets one build's `unregister()` null out another's still-live reference, leaking that listener for the rest of the daemon's life.

**Minor/cleanup (verified, low severity):**
5. `lib/index.ts` `buildArgs` calls `supportsConfigurationCache(gradleVersion)` three times for the same input — worth hoisting to a local.
6. `lib/init-cc.gradle`'s `wire()` stringifies every matching configuration's attributes twice when `-PconfAttr` is set (once in the filter, once for `observedAttributes`).

Ran a set of additional angles (cross-file call-site tracing including a real `npm install`/`tsc`/`jest` run, and a root-cause pass on the version-detection split across `index.ts`/`init-cc.gradle`) that turned up nothing else rising to a postable finding — the resolvableConfigurations/resolvedNames edge cases some sub-agents flagged turned out to be pre-existing behavior already present in `lib/init.gradle`, not introduced here.

---
_Generated by [Claude Code](https://claude.ai/code)_

lib/index.ts
trevjonez: Confirmed, but the deeper fix rather than bumping the floor to 8.2: Gradle DOES print the patch when it's nonzero - verified `gradle -v` on 8.10.1 prints "Gradle 8.10.1", and on 8.13.0 prints "Gradle 8.13" (patch omitted only because it's zero). The regex just wasn't capturing it. Fixed `parseGradleVersion` to capture major/minor/patch and gate on the real 8.1.1 floor, so 8.1.0 correctly stays on `init.gradle` and 8.1.1+ correctly gets `init-cc.gradle` - rather than excluding legitimate 8.1.1+ users along with 8.1.0. Pushed in 085b60f.

lib/init-cc.gradle
trevjonez: Confirmed. Restructured `SnykResourceUrlIndex` so the filename->URL map is no longer a class-level static at all - it's scoped inside a per-Gradle-instance `Registration` object, discarded when `unregister()` removes that instance's entry. Same change also fixes the listener-scoping issue in the next comment, since both were symptoms of the same "one shared static for what should be per-build state" design. Pushed in 085b60f.

lib/init-cc.gradle
trevjonez: Confirmed. Moved the gate from the init script's top level (where only `gradle.startParameter.projectProperties`, CLI-only, is available) into `SnykProjectConfigurer.configure()`, using the identical `gradleProperty`-backed read `readScanSpec` already uses. Verified end-to-end: a fixture with `snykIncludeComponentMetadata=true` set only in `gradle.properties` (no `-P`) now emits `distribution:url`, where before this fix it silently didn't. Pushed in 085b60f.

lib/init-cc.gradle
trevjonez: Confirmed. `SnykResourceUrlIndex` is now keyed by `System.identityHashCode(gradle)` per Gradle instance rather than a single overwritable static slot, so a `buildSrc`/included-build registration and unregistration can no longer disturb the root build's own listener. `register()` is idempotent per instance (`computeIfAbsent`) since every project now calls it from `configure()`; `unregister(int)` removes exactly one instance's entry. The identity is threaded through as a plain `int` - through `SnykCollectorParameters`, `SnykGraphBuilder`, and the task's `fragment` provider - specifically because a live `Gradle`/`Project` reference can't cross the isolated-action or configuration-cache boundary, which is the constraint the rest of this file is built around. Pushed in 085b60f.

One residual gap remained even after that fix, caught by Cursor Bugbot below: `buildSrc`/included builds never actually run the collector task, so the service backing `close()` is never instantiated there and `unregister()` never fires for that instance either. See the reply on that thread for the fix (085b60f -> 6b86619).

lib/index.ts
trevjonez: Done, hoisted into `const supportsCC` at the top of `buildArgs`, all three call sites now reference it. Pushed in 085b60f.

lib/init-cc.gradle
trevjonez: Done. `matchesAttributeFilter` now takes a pre-built snapshot (`Map<String,String>` of attribute name -> stringified value) instead of the `Configuration` itself; `wire()` builds that snapshot once per name-matching configuration into `attributesByConfig` and reads it for both the `-PconfAttr` decision and `observedAttributes`, so `getAttribute()` - the call that can realize a lazy/provider-backed attribute - only runs once per attribute per configuration instead of twice. Pushed in 085b60f.

lib/init-cc.gradle
```diff
@@ -0,0 +1,1236 @@
...
+        try {
+            Registration registration = REGISTRATIONS.remove(gradleIdentity)
+            if (registration?.listenerManager != null && registration.listener != null) {
+                registration.listenerManager.removeListener(registration.listener)
+            }
+        } catch (Exception ignored) {
+            // The listener is reclaimed when the daemon exits either way; removing
+            // the map entry above already drops this instance's captured URLs.
+        }
+    }
```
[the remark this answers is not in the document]
trevjonez: Confirmed. The identity-scoped fix in 085b60f corrected which instance's listener gets removed, but didn't address whether it ever gets removed at all for buildSrc/included builds - they never run snykResolvedDepsJson, so the SnykDepsCollector backing close() is never instantiated there and unregister() never fires.

Considered and rejected two ways to make cleanup itself more reliable rather than not registering in the first place:
- gradle.buildFinished, scoped to just the registering instance. Verified directly: Gradle refuses to store a configuration cache entry at all if anything registers a Gradle.buildFinished listener, full stop - "registration of listener on 'Gradle.buildFinished' is unsupported" - regardless of what the listener closes over.
- Forcing eager instantiation of the collector service so Gradle's own service-close mechanism fires unconditionally. Worse bug: close() unconditionally prints a JSONDEPS/JSONATTRS line, and extractJsonFromScriptOutput keeps the FIRST JSONDEPS line it sees. buildSrc finishes before the root build's own scan task runs, so its close() would print first and the real result would be discarded.

Fixed by not registering there at all, since it was always pure waste - gated on `project.gradle.parent == null`, which is null for the build actually being scanned and non-null for buildSrc/included builds (verified directly: root parent=null, buildSrc parent=build 'init-gradle', distinct identityHashCode values). Pushed in 6b86619.

PR #356, opened 2026-09-17
fix: [OSM-3962] stop re-walking every route when building the verbose dep graph

jsutherland-snyk:
One fix to the Gradle verbose dependency-graph builder. snyk/snyk-mvn-plugin#237 is the Maven half, same rule and same helpers; review them together. Either can land first. [OSM-3962](https://snyksec.atlassian.net/browse/OSM-3962)

**The verbose walk decides each edge once instead of re-queueing every route.** Re-queueing a visited package's children per incoming route made construction O(routes): 28 packages / 81 edges took 39,551ms, all after Gradle had finished. `--print-graph` sets `verbose`, so every Gradle SBOM takes this path.

**Cycles are decided on resolved ids, by component membership and per-cycle dominance.** `sha1Map` can collapse two raw ids onto one coordinate, which made a package depend on itself. Dominator trees are built only inside cyclic components, so acyclic packages skip dominance entirely.

How to verify: `npx jest test/functional` - 18 tests in `graph.spec.ts`, including a chain plus shared-leaf case that fails on the old code. 128,000 packages build in about 2s.

Caveats: that shape inside one large cycle is still quadratic (2.2s at 8,000 packages). Extra `pruned: cyclic` placeholders can appear; no package or real edge is lost. Tarjan and the dominator helpers are duplicated with Maven and belong in `@snyk/dep-graph`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Changes SBOM/verbose dependency output semantics and performance for all Gradle scans using `--print-graph`, with complex graph logic that can differ slightly from the old per-route behavior despite strong test coverage.
> 
> **Overview**
> **Replaces the verbose Gradle dependency-graph builder** (`--print-graph`) so it no longer walks every distinct path through the graph. The old BFS re-queued children per incoming route (O(routes), minutes on large reactors); verbose mode now uses a dedicated `buildVerboseGraph` that decides each edge once.
> 
> **Cycle handling uses graph analysis instead of per-path ancestry.** Strongly connected components (Tarjan) and dominator trees scoped to cyclic components decide when to emit `:pruned` nodes with `pruned: cyclic` versus keeping the plain dependency edge. Cycle logic runs in **sha1-resolved id space** so collapsed coordinates do not become self-dependencies; edge cases (root id collision, null names) are guarded to match the non-verbose path.
> 
> **Non-verbose `buildGraph` is simplified:** ancestry tracking and verbose-only revisit logic are removed from the shared BFS; revisits always get a `pruned: true` placeholder as before.
> 
> **Semantics caveat:** cycle-edge detection is an intentional over-approximation (may add extra cyclic placeholders; packages and real edges are not dropped). Functional tests add correctness pins and wall-clock regression guards for fan-in graphs and long chains with a shared library.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit af02e3f79ebc27e23e52b2ac1d17fbd10c7325df. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

[OSM-3962]: https://snyksec.atlassian.net/browse/OSM-3962?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/graph.ts
```diff
@@ -144,16 +131,427 @@ export async function buildGraph(
...
+    return resolvedId;
+  };
+
+  // Raw ids that resolve to one coordinate have to agree on the metadata they
+  // contribute. The old walk took whichever the traversal happened to reach
+  // first; picking the lowest raw id instead is arbitrary in the same way but
+  // stable, so the emitted graph does not depend on traversal order.
+  //
+  // It has to be the lowest raw id the walk could actually have *reached*,
+  // though. An unreachable group member contributes nothing to the graph, and
```
calhar-snyk, line 186: I feel like we're doing a lot of reachability work here that is not necessarily justified? Can we actually produce a gradle graph with nodes that are not reachable from the root?

lib/graph.ts
```diff
@@ -144,16 +131,427 @@ export async function buildGraph(
...
+  // letting it win hands the package that member's hashes and distributionUrl
+  // - so a component's sha1 comes out wrong and its distributionUrl goes
+  // missing, which is exactly what an SBOM reports.
+  const rawIdsByResolvedId = new Map<string, string[]>();
+  for (const rawId of Object.keys(gradleGraph)) {
```
calhar-snyk, line 191: Nitpick: If we want to drop node20 support -- this becomes the much nicer:
```
  const rawIdsByResolvedId = Map.groupBy(
    Object.keys(gradleGraph),
    resolvedIdOf,
  );
```

lib/graph.ts
```diff
@@ -144,16 +131,427 @@ export async function buildGraph(
...
+    rawIds.sort((a, b) => {
+      const aReached = rawReachable.has(a);
+      if (aReached !== rawReachable.has(b)) return aReached ? -1 : 1;
+      return a < b ? -1 : a > b ? 1 : 0;
+    });
+  }
+
+  const childrenOf = (resolvedId: string): string[] => {
+    const children: string[] = [];
+    for (const rawId of rawIdsByResolvedId.get(resolvedId) || [resolvedId]) {
```
calhar-snyk, line 217: This safety fallback seems unnecessary? Given the pre-processing -- the only time we should get a null value here is for the root node?

lib/graph.ts
```diff
@@ -144,16 +131,427 @@ export async function buildGraph(
...
+// A node id the graph cannot contain, so the dominator tree can have an entry
+// of its own that is distinct from the dependency graph's root.
+const DOMINANCE_ENTRY = '\u0000dominance-entry';
+
+// `to` is reachable from the root without `from` exactly when `from` does not
+// dominate `to`, so one dominator tree answers every such question in constant
+// time. Answering them with a reachability pass per cycle member instead cost
+// O(cycle members x nodes) in both time and memory, which is worse than the
+// walk it replaced on a graph made of many small cycles.
+function buildDominanceTest(
```
calhar-snyk, line 366: I think this function as a whole could be improved. I get that the world is doing the whole "only the LLM needs to actually understand the code, human maintainability is dead". But this can surely be split into some sensible divisions.

lib/graph.ts
```diff
@@ -144,16 +131,427 @@ export async function buildGraph(
...
+    for (const child of childrenIn(id)) {
+      if (!rank.has(child)) continue;
+      const known = predecessors.get(child);
+      if (known) known.push(id);
+      else predecessors.set(child, [id]);
+    }
+  }
+
+  // Cooper, Harvey and Kennedy's iterative formulation
+  const idom = new Map<string, string>([[rootId, rootId]]);
```
calhar-snyk, line 406: While conventional in terms of the algorithm. Not obvious this is for the immediate dominator. Especially when we have `commonDominator` as a function.

lib/graph.ts
calhar-snyk: Broadly, I think there are three main pieces here that can be sensibly split out into their own functions with improved descriptiveness:

- `prepareDominanceGraph` -- building the `order`, `rank`, `predecessors` objects.
- `computeImmediateDominators` -- producing the idom map and iterating until settled.
- `createDominanceLookup` -- Constructing the `entered`/`exit`  maps and creating the dominance function.

lib/graph.ts
```diff
@@ -144,16 +131,427 @@ export async function buildGraph(
...
+    return (
+      from <= to &&
+      (exited.get(id) as number) <= (exited.get(dominator) as number)
+    );
+  };
+}
+
+// Tarjan's algorithm, driven by an explicit stack: a recursive implementation
+// overflows the call stack on the deep dependency chains this has to cope with
+// (the same trap as CMPA-770 in lib/init.gradle).
```
calhar-snyk, line 484: Nitpick: Ticket IDs in comments should only be for planned/future work

lib/graph.ts
```diff
@@ -144,16 +131,427 @@ export async function buildGraph(
...
+    !dominates(to, from);
+
+  type Coordinates = {
+    name: string;
+    version: string;
+    pkgIdProvenance?: string;
+    hashes?: Record<string, string>;
+    distributionUrl?: string;
+  };
+  const coordinatesCache = new Map<string, Coordinates>();
```
calhar-snyk, line 275: Building this cache and the `coordinatesOf` function could also be split out I think. It's just bloating the  function

lib/graph.ts
```diff
@@ -144,16 +131,427 @@ export async function buildGraph(
...
+          added.add(prunedId);
+          const coordinates = coordinatesOf(to);
+          depGraphBuilder.addPkgNode(
+            { name: coordinates.name, version: coordinates.version },
+            prunedId,
+            createNodeInfo(coordinates.pkgIdProvenance, 'cyclic'),
+          );
+        }
+        depGraphBuilder.connectDep(from, prunedId);
+        if (!routeAvoiding(from, to)) continue;
```
calhar-snyk, line 348: Nitpick: The naming here is clear in theory. But the mechanism -- particularly knowing that this is really `!!dominates(to, from)` so `dominates(to, from)` make this really weird to think about.

calhar-snyk: I think either the `routesAvoiding` function or here needs a bit of an explanatory comment. Possibly a clearer name `canReachSourceWithoutTarget` or something to that effect.

lib/graph.ts
```diff
@@ -144,16 +131,427 @@ export async function buildGraph(
...
+  childrenOf: (id: string) => string[],
+): (dominator: string, id: string) => boolean {
+  const childrenIn = (id: string): string[] =>
+    childrenOf(id).filter((child) => reachable.has(child));
+
+  // depth-first postorder, then reversed, so every node follows its
+  // predecessors wherever the graph is acyclic
+  const postorder: string[] = [];
+  const seen = new Set<string>([rootId]);
+  const dfs = [{ id: rootId, children: childrenIn(rootId), next: 0 }];
```
calhar-snyk, line 378: I'm being quite nitpicky here. But I find this particularly construction of the iterative DFS a lot harder to parse than the one from snyk/cli-extension-dep-graph#255 which has a leaving and a regular frame. I understand the special case for Tarjan's given that each frame provides additional context.
But I think this one and the `treeChildren` walk can be simplified.

lib/graph.ts
jsutherland-snyk: No - init.gradle only adds a node together with an edge from the root or from an already-added node, so everything is reachable. The reachability passes, the reachable-first metadata ordering and the orphan-node test are gone in c0d2283.

lib/graph.ts
jsutherland-snyk: Agreed it reads better, but `engines` still allows Node 20, so I have left it for now. It is pulled out into `groupRawIdsByResolvedId`, so it is a one-line swap when we drop 20.

lib/graph.ts
jsutherland-snyk: Agreed, removed in c0d2283.

lib/graph.ts
jsutherland-snyk: Fair. Split in c0d2283 into `prepareDominanceGraph`, `computeImmediateDominators` and `createDominanceLookup`, with `buildCycleDominanceTest` wiring them together.

lib/graph.ts
jsutherland-snyk: Renamed to `immediateDominator` in c0d2283, and `commonDominator` is now `nearestCommonDominator`.

lib/graph.ts
jsutherland-snyk: Removed in c0d2283.

lib/graph.ts
jsutherland-snyk: Extracted into `createCoordinatesLookup` in c0d2283.

lib/graph.ts
jsutherland-snyk: Agreed, renamed to `canReachSourceWithoutTarget` in c0d2283, which says what it checks rather than the negation.

lib/graph.ts
jsutherland-snyk: Both walks now use a leaving frame, same shape as snyk/cli-extension-dep-graph#255. Done in c0d2283.

calhar-snyk (approved):

